PR #1, opened 2020-02-13
Update README.md

roshray:
Talks on HE.


youben11:
Thanks @roshray and sorry for the late reply. The tutorial section is more about how to use TenSEAL which should be added after TenSEAL API is usable.

PR #3, opened 2020-02-22
Setup the build process

chinmayshah99:
- Added instructions in README to start building.
- Updated Cmakelists, making sure build is happening. 
- added a basic test file to verify build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,13 @@ TenSEAL is a library for doing homomorphic encryption operations on tensors, bui
 Install CMake. 
 If you are on Linux, head down to [CMake- Snap](https://snapcraft.io/cmake) and install CMake.
 
-Run ```prereqs``` to setup dependecies.
+For people on Linux, Run ```prereqs``` to setup dependecies.
+
+For other platform:
+```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -4,7 +4,12 @@ TenSEAL is a library for doing homomorphic encryption operations on tensors, bui
 
 ## Installation
 
-TODO
+Install CMake. 
+If you are on Linux, head down to [CMake- Snap](https://snapcraft.io/cmake) and install CMake.
+
+Run ```prereqs``` to setup dependecies.
```
youben11: I think we should explain what the user should do here (get those third party library) instead of providing a bash script that's somehow platform dependent.

chinmayshah99: What can be one approach is use Docker for development, so development is only done on Linux. So we don't need to worry about supporting anything else. 

But surely, I'll add the instructions just in case for non Linux development.

chinmayshah99: Done!

youben11: I'm working out a Dockerfile which should be available today.

PR #5, opened 2020-03-29
BFV naive vector type

youben11:
This PR introduces some functions for key generation of the BFV scheme, namely, the creation of a SEAL parameter and context object that are then used to generate a public and secret key directly from python.

It also introduces the BFVNaiveVector class, which can hold an encrypted vector of integers, we can add, multiply and sub BFVNaiveVector objects, as well as add, multiply and substract them with plaintext vectors (list of int). Here I don't encode the vector of integers into a single ciphertext, but rather use a ciphertext for each integer (that's why I used the term "naive").


philomath213:
@youben11 you forget to initialize the _i_ variable in _for_ loop inside:
_BFVNaiveVector_ constructor `for (int i; i < vec.size(); i++) {`
_BFVNaiveVector::decrypt_ function `for (int i; i < this->ciphertexts.size(); i++) {`
This may lead to undefined behavior.

PR #8, opened 2020-04-03
TenSEALContext: store keys and parameters on a single location

youben11:
On a previous PR #7  @s0l0ist suggested to solve the issue of instantiating some of the objects required by SEAL (e.g. Encryptor) on every time we need them.

This PR introduce TenSEALContext class, which is basically a way of keeping a context which includes keys and parameters used during evaluation, encryption and decryption, as well as other objects required during evaluation like the Evaluator object. A single TenSEALContext is intended to be shared across multiple vectors, they will only hold a pointer to it.

This change simplified both the code and the API, many thanks to @s0l0ist :clap:


tenseal/tensors/__init__.py
```diff
@@ -6,39 +6,37 @@
...
-def bfv_naive_vector(context, public_key, plaintext_vector):
+def bfv_naive_vector(context, plaintext_vector):
     """Constructor method for the BFVNaiveVector object, which can store
     a list of integers in encrypted form, using the BFV homomorphic encryption
     scheme.
 
     Args:
-        context: a SEALContext object, holding the encryption parameters.
-        public_key: a PublicKey object to be used for encrypting the vector.
+        context: a BFVContext object, holding the encryption parameters and keys.
```
philomath213, line 15: doc update: now it's TenSEALContext object

tenseal/tensors/__init__.py
```diff
@@ -6,39 +6,37 @@
...
+def ckks_vector(context, scale, plaintext_vector):
     """Constructor method for the CKKSVector object, which can store a list
-    of integers in encrypted form, using the CKKS homomorphic encryption
+    of float numbers in encrypted form, using the CKKS homomorphic encryption
     scheme.
 
     Args:
-        context: a SEALContext object, holding the encryption parameters.
-        public_key: a PublicKey object to be used for encrypting the vector.
+        context: a CKKSContext object, holding the encryption parameters and keys.
```
philomath213, line 31: doc update: now it's TenSEALContext object instead of CKKSContext

PR #15, opened 2020-04-09
Flexible TenSEALContext

philomath213:
This PR modify the TenSEALContext class and its behaviors according to Keys Generations.
The KeyGenerator object is no longer an attribute, instead there's a pointer for each key (Public, Secret, Galois and Relinearization keys).
During constructing a TenSEALContext object we create a KeyGenerator object in order to generate a public and secret keys and making encrypter and decrypter.
There's a new method `make_context_public`, it will generate the Galois and Relinearization keys if needed, then make the context public, by destroying the secret key and the decryptor and setting them to `nullptr`.

### DIFF
--- a/tenseal/tensealcontext.h
+++ b/tenseal/tensealcontext.h
@@ -62,11 +62,62 @@ class TenSEALContext {
         return shared_ptr<TenSEALContext>(new TenSEALContext(filename));
     }
 
-    shared_ptr<KeyGenerator> key_gen() { return this->_keygen; }
-    PublicKey public_key() { return this->_keygen->public_key(); }
-    SecretKey secret_key() { return this->_keygen->secret_key(); }
-    RelinKeys relin_keys() { return this->_keygen->relin_keys(); }
-    GaloisKeys galois_keys() { return this->_keygen->galois_keys(); }
+    PublicKey public_key() { return PublicKey(*this->_public_key); }
+    SecretKey secret_key() {
+        if (this->_secret_key == nullptr) {
+            throw invalid_argument(
+                "the current context is public, it doesn't hold a Secret key");
+        }
+
+        return SecretKey(*this->_secret_key);
+    }
+
+    RelinKeys relin_keys() {
+        if (this->_relin_keys == nullptr) {
+            throw invalid_argument(
+                "the current context doesn't hold a Relinearization keys");
+        }
+
+        return RelinKeys(*this->_relin_keys);
+    }
+
+    GaloisKeys galois_keys() {
+        if (this->_galois_keys == nullptr) {
+            throw invalid_argument(
+                "the current context doesn't hold a Galois keys");
+        }
+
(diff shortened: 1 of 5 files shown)


tenseal/tensealcontext.h
```diff
@@ -62,11 +62,63 @@ class TenSEALContext {
...
+        }
+
+        // destory and set _secret_key and decryptor to null
+        this->_secret_key = nullptr;
+        this->decryptor = nullptr;
+        this->_is_public = true;
+    }
 
     bool is_public() { return this->_is_public; }
     bool is_private() { return !this->_is_public; }
```
youben11, line 124: I think we can now get rid of the _is_public attribute and base these functions on either _secret_key in nullptr

tenseal/tensealcontext.h
```diff
@@ -62,11 +62,63 @@ class TenSEALContext {
...
+        // create KeyGenerator object only if needed
+        if (generate_galois_keys || generate_relin_keys) {
+            KeyGenerator keygen = KeyGenerator(
+                this->_context, *this->_secret_key, *this->_public_key);
+
+            // generate Galois Keys
+            if (generate_galois_keys) {
+                this->_galois_keys = shared_ptr<GaloisKeys>(
+                    new GaloisKeys(keygen.galois_keys()));
+            }
```
youben11, line 108: I think this should be a separate method and can be called outside of make_context_public as well as a secret and public key, or a key_generator (like in this case)

tenseal/tensealcontext.h
```diff
@@ -62,11 +62,63 @@ class TenSEALContext {
...
+
+            // generate Relinearization Keys
+            if (generate_relin_keys) {
+                this->_relin_keys =
+                    shared_ptr<RelinKeys>(new RelinKeys(keygen.relin_keys()));
+            }
+        }
```
youben11, line 115: same as for galois keys

tenseal/tensealcontext.h
```diff
@@ -62,11 +62,63 @@ class TenSEALContext {
...
+                "the current context is public, it doesn't hold a Secret key");
+        }
+
+        return SecretKey(*this->_secret_key);
+    }
+
+    RelinKeys relin_keys() {
+        if (this->_relin_keys == nullptr) {
+            throw invalid_argument(
+                "the current context doesn't hold a Relinearization keys");
```
youben11, line 78: Can hint to call generate_relin_keys()

tenseal/tensealcontext.h
```diff
@@ -62,11 +62,63 @@ class TenSEALContext {
...
+        }
+
+        return RelinKeys(*this->_relin_keys);
+    }
+
+    GaloisKeys galois_keys() {
+        if (this->_galois_keys == nullptr) {
+            throw invalid_argument(
+                "the current context doesn't hold a Galois keys");
```
youben11, line 87: Can hint to call generate_galois_keys()

PR #18, opened 2020-04-10
Set up `scikit-build` for packaging and distribution

karlhigley:
Trying to get this set up to work on macOS at least, and hoping we can use [`dockcross`](https://github.com/dockcross/dockcross) to build wheels for other platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,11 @@
 cmake_minimum_required(VERSION 3.12)
 
+if(SKBUILD)
+  message(STATUS "The project is being built using scikit-build")
+else()
+  message(STATUS "The project is NOT being built using scikit-build")
+endif()
+
 project(tenseal)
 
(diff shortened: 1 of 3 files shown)


youben11:
Why do we need scikit-build to use it with dockross? if the docker images are ready then using `pip` and `auditwheel` should be enough I think to get the wheels ready to be pushed

karlhigley:
It has [good support](https://scikit-build.readthedocs.io/en/latest/generators.html) for a variety of compiler toolchains without having to write much code in `setup.py`.

karlhigley:
(And should work with `manylinux` too.)

youben11:
Thanks a lot @karlhigley :heart: but we preferred to keep the same build tools and automate the publishing on release for Linux #27 and MacOS #25 (Windows is on its way) via github actions.

karlhigley:
Excellent! Let's close this then.

PR #21, opened 2020-04-23
SEAL API

### LINKED ISSUE #10: Provide SEAL API
### Context
The goal of TenSEAL isn't to wrap SEAL and make it available to use in Python but as we have everything setup to make a full port of it to Python (making every cpp object and API call available through python), I think that it would be nice to make a tenseal.seal_api submodule that do this.

### TODO
- Bind all SEAL objects (providing access to all their methods and public attributes), preferably done in a separate file `sealapi.cpp` under `tenseal/`.
- Customize the build (`setup.py` and `CMakeLists.txt`) to make the build of the SEAL API optional.

bcebere:
Expose the SEAL API in the library

Fixes #10


tenseal/sealapi.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include <pybind11/operators.h>
+#include <pybind11/pybind11.h>
+#include <seal/seal.h>
+
+using namespace seal;
+using namespace std;
+namespace py = pybind11;
+
+void loadSEALAPI(py::module &m) {
```
youben11, line 11: seal_api should be totally independent, it can be done like this
```suggestion
PYBIND11_MODULE(_tenseal_cpp, m) {
    py::module seal_api = m.def_submodule("seal");
    // "seal/biguint.h"
    py::class_<BigUInt>(seal_api, "BigUInt")
```
This way, you won't need to call it from binding.cpp

tenseal/binding.cpp
```diff
@@ -194,4 +195,6 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
         .value("NONE", scheme_type::none)
         .value("BFV", scheme_type::BFV)
         .value("CKKS", scheme_type::CKKS);
+
+    loadSEALAPI(m);
```
youben11, line 199: We shouldn't know about seal_api here

tenseal/binding.cpp
```diff
@@ -5,6 +5,7 @@
 #include <memory>
 #include <vector>
 
+#include "sealapi.h"
```
youben11, line 8: we shouldn't know about sealapi here

tenseal/__init__.py
```diff
@@ -11,6 +11,7 @@
 SecretKey = _ts_cpp.SecretKey
 RelinKeys = _ts_cpp.RelinKeys
 GaloisKeys = _ts_cpp.GaloisKeys
+BigUInt = _ts_cpp.BigUInt
```
youben11, line 14: Everything that relates to seal_api should be under `tenseal.seal_api` namespace and not directly in `tenseal`

CMakeLists.txt
```diff
@@ -11,12 +11,15 @@ add_subdirectory(third_party/pybind11)
...
     ${TENSEAL_BASEDIR}/binding.cpp
+    ${TENSEAL_BASEDIR}/sealapi.cpp
     ${TENSEAL_BASEDIR}/tensealcontext.cpp
     ${TENSEAL_BASEDIR}/utils.cpp
     ${TENSEAL_BASEDIR}/tensors/bfvnaivevector.cpp
     ${TENSEAL_BASEDIR}/tensors/bfvvector.cpp
     ${TENSEAL_BASEDIR}/tensors/ckksvector.cpp
 )
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wnarrowing -Wextra -Wno-varargs -Werror -Wno-strict-overflow -Wfatal-errors -fno-strict-overflow -Wno-psabi -Wno-noexcept-type -Wno-maybe-uninitialized -fstack-protector")
```
youben11, line 22: Wouldn't this stop the build if those flags aren't available on the target platform? I would maybe suggest adding them if available

CMakeLists.txt
```diff
@@ -11,12 +11,15 @@ add_subdirectory(third_party/pybind11)
 set(TENSEAL_BASEDIR tenseal)
 set(SOURCES
     ${TENSEAL_BASEDIR}/binding.cpp
+    ${TENSEAL_BASEDIR}/sealapi.cpp
```
youben11, line 14: Not necessary for now, but this line should be made optional, so that we would compile seal_api into TenSEAL or not.

tests/test_sealapi.py
```diff
@@ -0,0 +1,442 @@
...
+    ]
+
+    for scheme_id, scheme in enumerate(schemes):
+        assert int(scheme) == scheme_id
+
+        testcase = sealapi.EncryptionParameters(scheme)
+        assert testcase.scheme() == scheme
+
+        testcase = sealapi.EncryptionParameters(scheme_id)
+        assert testcase.scheme() == scheme
```
youben11, line 265: You can use pytest.mark.parametrize when you need to do similar test on different values.

tests/test_sealapi.py
```diff
@@ -0,0 +1,442 @@
...
+
+    testcase = sealapi.EncryptionParameters(sealapi.SCHEME_TYPE.NONE)
+    poly_modulus_degree_failed = False
+    try:
+        testcase.set_poly_modulus_degree(10)
+    except:
+        poly_modulus_degree_failed = True
```
youben11, line 272: Can use pytest.raises for checking if an exception was raised

tenseal/sealapi/sealapi.cpp
```diff
@@ -0,0 +1,376 @@
...
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <seal/seal.h>
+
+using namespace seal;
+using namespace std;
+namespace py = pybind11;
+
+PYBIND11_MODULE(_sealapi_cpp, m) {
+    m.doc() = "Library for doing homomorphic encryption operations";
```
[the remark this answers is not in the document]
bcebere: I will add that, thank you.

Speaking of docs, where would it be best to explain what objects and methods I exported from SEAL? I will leave the full explanation for their headers, they are doing an excelent job.

But for those importing tenseal. Should I add some docstrings to sealapi's __init__?

tests/test_sealapi.py
```diff
@@ -0,0 +1,442 @@
...
+    header = sealapi.Serialization.SEALHeader()
+    assert header.magic == 0xA15E
+    assert header.zero_byte == 0
+    assert header.compr_mode == sealapi.COMPR_MODE_TYPE.NONE
+    assert header.size == 0
+    assert header.reserved == 0
+
+    assert sealapi.Serialization.IsSupportedComprMode(
+        sealapi.COMPR_MODE_TYPE.NONE) == True
+    assert sealapi.Serialization.IsSupportedComprMode(15) == False
```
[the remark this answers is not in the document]
bcebere: I just wanting force the function to return False.

I am still getting familiar with SEAL and pybind11, I will do a second sweep through the headers and I'll review what makes sense or not(in terms of exporting or testing).

tenseal/sealapi/sealapi.cpp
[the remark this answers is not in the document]
youben11: The best docstrings to put there in my POV is to specify that it's a full port of the SEAL API and whoever wants to learn about a specific function should check SEAL API documentation.

tests/test_sealapi.py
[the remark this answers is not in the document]
youben11: Okay, just wanted to make sure it's not the value for the zlib compression mode.

bcebere:
**Issue** 
no support for updating  std::vector references, we must return the value instead. One extra copy involved.
 
They recommend making the std::vector container opaque, but that breaks the vector binding entirely.

For example, BatchEncoder method decode
`void decode(const Plaintext &plain, std::vector<std::uint64_t> &destination)`
becomes
`std::vector<std::uint64_t> decode_uint64(const Plaintext &plain)`
Ref: https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html

tenseal/sealapi/sealapi_evaluator.cpp
```diff
@@ -0,0 +1,302 @@
...
+     ***/
+    py::class_<Evaluator, std::shared_ptr<Evaluator>>(m, "Evaluator",
+                                                      py::module_local())
+        .def(py::init<std::shared_ptr<SEALContext>>())
+        .def("add_inplace", &Evaluator::add_inplace)
+        .def("sub_inplace", &Evaluator::sub_inplace)
+        .def("multiply_inplace",
+             [](Evaluator &e, Ciphertext &l, const Ciphertext &r) {
+                 return e.multiply_inplace(l, r);
+             })
```
youben11, line 28: Q: Is this and similar inplace functions not doing it really inplace?

CMakeLists.txt
```diff
@@ -17,6 +17,23 @@ set(SOURCES
...
+    ${TENSEAL_BASEDIR}/sealapi/sealapi_context.cpp
+)
+pybind11_add_module(_sealapi_cpp ${SEALAPI_SOURCES})
+target_link_libraries(_sealapi_cpp PRIVATE seal)
+
+set(SEALAPI_UTIL_SOURCES
+    ${TENSEAL_BASEDIR}/sealapi/util/util.cpp
+)
+pybind11_add_module(_sealapi_util_cpp ${SEALAPI_UTIL_SOURCES})
+target_link_libraries(_sealapi_util_cpp PRIVATE seal)
```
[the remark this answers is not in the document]
bcebere: Just for importing it as sealapi.util.
I will merge the libraries

tenseal/sealapi/sealapi_evaluator.cpp
bcebere: return does nothing here.
I will update these bindings with the py::arg() format.

bcebere: I removed the useless return keywords.

tenseal/sealapi/sealapi_encode.cpp
```diff
@@ -11,6 +11,13 @@ using namespace seal;
 using namespace std;
 namespace py = pybind11;
 
+/***
+ *Notes:
+ * Some methods, like multiply_inplace, have dedicated lambda implementation
```
youben11, line 16: I think you meant decode here?

bcebere: yes, my bad

Issue #35, opened 2020-05-04
Relinearization [and Rescaling]


youben11:
After every multiplication for CKKS and BFV, we need to relinearize our ciphertext, CKKS also need rescaling (as the scale becomes Delta**2). We need to have these operations done automatically by default after every multiplication, with the ability to choose to not do so.

RafiulIslamRafi:
Which function gives Relinearization in TenSEAL? when I multiply a variable for 2nd time in cyphertext, it returns a garbage result.

PR #42, opened 2020-05-09
BFV vectors relinearization after ct-ct multiplication

philomath213:
Relinearize BFV vectors after ct-ct multiplication.

Related #35

### DIFF
--- a/tenseal/tensors/bfvvector.cpp
+++ b/tenseal/tensors/bfvvector.cpp
@@ -171,6 +171,11 @@ BFVVector& BFVVector::mul_inplace(BFVVector to_mul) {
 
     this->context->evaluator->multiply_inplace(this->ciphertext,
                                                to_mul.ciphertext);
+                                               
+    // TODO: make this optional
+    // relineraization after ciphertext-ciphertext multiplication
+    this->context->evaluator->relinearize_inplace(this->ciphertext,
+                                                  this->context->relin_keys());
 
     return *this;
 }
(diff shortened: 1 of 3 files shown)


tenseal/tensors/bfvvector.cpp
```diff
@@ -171,6 +171,11 @@ BFVVector& BFVVector::mul_inplace(BFVVector to_mul) {
 
     this->context->evaluator->multiply_inplace(this->ciphertext,
                                                to_mul.ciphertext);
+                                               
+    // TODO: make this optional
+    // relineraization after ciphertext-ciphertext multiplication
+    this->context->evaluator->relinearize_inplace(this->ciphertext,
+                                                  this->context->relin_keys());
```
bcebere, line 178: Maybe it would be nice to add some tests that verify the size of the ciphertext after the bfvvector multiplication, to underline the importance of the fix.

youben11:
I guess you just need to do some linting

tenseal/tensors/bfvvector.cpp
bcebere, line 178: Wouldn't be useful to support relinearization after a batch of operations? 
Meaning:
- you could let the user choose when to relinearize.
 - or you could let them do a set of operations in a row that have at least a ct-ct multiplication, and just after them to relinearize.

youben11: I love this, we could add for both BFV and CKKS!

youben11: We aim at providing the ability to the user to disable relinearization in the near future so that he can do that himself, but the main rule for the moment is to relinearize after each mul to keep the ciphertext at its smallest size

philomath213: Yeah, I think we should add these tests.

PR #45, opened 2020-05-10
Tenseal encoders

### LINKED ISSUE #43: Make encoders part of TenSEALContext
### Context

Current implementation instantiate an encoder whenever needed which might incur some extra overhead, and reserving space for all the 3 different kind of encoders is non-efficient.

### Possible solution

Keep 3 pointers to all kind of encoders in `TenSEALContext` (`[Integer|Batch]Encoder` for BFV and `CKKSEncoder` for CKKS) and have a method `get_encoder` that when asked for a specific encoder, check either it's created or not yet, create a new one if it's not, and return a pointer to the encoder.

**Question to take into account:** is it safe to share pointers to encoders? are they thread-safe?

bcebere:
- Add encoders to Tenseal context.
- Add some basic benchmarks for profiling the changes.

Fixes #43

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -3,3 +3,4 @@ bumpversion>=0.5.3
 flake8
 pytest>=5.4.1
 setuptools>=41.0.0
+pytest-benchmark
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -4,3 +4,4 @@ flake8
 pytest>=5.4.1
 setuptools>=41.0.0
 pytest-benchmark
+pytest-benchmark[histogram]
--- a/tenseal/tensealcontext.cpp
+++ b/tenseal/tensealcontext.cpp
@@ -2,14 +2,12 @@
 
 #include <seal/seal.h>
 
-#include <vector>
-
 #include "utils.h"
+
+namespace tenseal {
 
 using namespace seal;
 using namespace std;
-
-namespace tenseal {
 
 TenSEALContext::TenSEALContext(EncryptionParameters parms) {
     this->_parms = parms;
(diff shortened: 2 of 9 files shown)


tenseal/tensealcontext.cpp
```diff
@@ -2,30 +2,27 @@
...
 #include <seal/seal.h>
 
-#include <vector>
-
 #include "utils.h"
 
+namespace tenseal {
+
 using namespace seal;
 using namespace std;
```
youben11, line 10: Q: What's the advantage of putting this here?

tenseal/tensealcontext.cpp
```diff
@@ -2,30 +2,27 @@
...
-        new Decryptor(this->_context, *this->_secret_key));
-    this->evaluator = shared_ptr<Evaluator>(new Evaluator(this->_context));
+    this->_public_key = make_shared<PublicKey>(keygen.public_key());
+    this->_secret_key = make_shared<SecretKey>(keygen.secret_key());
+    this->encryptor =
+        make_shared<Encryptor>(this->_context, *this->_public_key);
+    this->decryptor =
+        make_shared<Decryptor>(this->_context, *this->_secret_key);
+    this->evaluator = make_shared<Evaluator>(this->_context);
+    this->encoder = make_shared<TenSEALEncoder>(this->_context);
```
youben11, line 25: I just learned about the advantages of using make_shared, and it looks even better

tenseal/tensealcontext.h
```diff
@@ -30,6 +29,7 @@ class TenSEALContext {
     shared_ptr<Encryptor> encryptor;
     shared_ptr<Decryptor> decryptor;
     shared_ptr<Evaluator> evaluator;
+    shared_ptr<TenSEALEncoder> encoder;
```
youben11, line 32: Would it make sense to have a name like encoder_factory, encoder_store or simply encoders? I think it would help differentiate with encoder objects.

philomath213: Same thoughts here, I would like to suggest `encoder_factory`.

requirements_dev.txt
```diff
@@ -3,3 +3,4 @@ bumpversion>=0.5.3
 flake8
 pytest>=5.4.1
 setuptools>=41.0.0
+pytest-benchmark
```
philomath213, line 6: It would be nice to have histograms using pytes-benchmark, this requires to install `pytest-benchmark[histogram]` instead.

tenseal/tensealcontext.cpp
bcebere: In the context of cpps, not much.

But it's usually for the best to avoid polluting the global namespace with "using" statements, or at least limit them to the scope of another namespace.
Another reason for using the in the cpp could be for preventing including headers after the using statement. It's usually recommended to add at least a anonymous namespace before adding "using".


In ".h" is not recommended at all, it must be scoped by a namespace, because you change the scope/API for all your integrators.

youben11: I know for header files, but thought that it doesn't affect for cpps

bcebere: The only issue with cpp files is adding an "using" statement before an "include" statement.
But in general, it can lead to annoying bugs anyway,

PR #55, opened 2020-05-14
Python publish workflow for Windows

### LINKED ISSUE #11: Automate build-push to PyPI
As TenSEAL includes a binary extension, it needs some extra attention on the build process to support different platforms, but the ideal would be to have this process automated on release, so whenever a new version comes in, it gets built and pushed to [PyPI](https://pypi.org/).

### Some resources

- https://github.com/pypa/manylinux
- https://packaging.python.org/tutorials/packaging-projects/
- https://packaging.python.org/guides/packaging-binary-extensions/
- https://packaging.python.org/specifications/platform-compatibility-tags/

philomath213:
Build and push wheels for Windows.

Fixes #11.

### DIFF
--- a/.github/workflows/pythonpublish-windows.yml
+++ b/.github/workflows/pythonpublish-windows.yml
@@ -3,8 +3,6 @@ name: Windows Package
 on:
   release:
     types: [created]
-  pull_request:
-    types: [ opened, synchronize, reopened ]
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


.github/workflows/pythonpublish-windows.yml
```diff
@@ -0,0 +1,40 @@
+name: Windows Package
+
+on:
+  release:
+    types: [created]
+  pull_request:
+    types: [ opened, synchronize, reopened ]
```
youben11, line 7: we just need to switch it to release only

PR #56, opened 2020-05-14
Testing two chanied matmul

youben11:
Add a single test function for testing two chained vector-matrix multiplciation

### DIFF
--- a/tests/tensors/test_ckks_vector.py
+++ b/tests/tensors/test_ckks_vector.py
@@ -427,7 +427,7 @@ def test_vec_plain_matrix_mul_depth2(context, scale, vec, matrix1, matrix2):
     ct = ts.ckks_vector(context, scale, vec)
     result = ct @ matrix1 @ matrix2
     expected = (np.array(vec) @ np.array(matrix1) @ np.array(matrix2)).tolist()
-    assert _almost_equal(result.decrypt(), expected, 1), "Matrix multiplciation is incorrect."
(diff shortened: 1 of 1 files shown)


tests/tensors/test_ckks_vector.py
```diff
@@ -1,5 +1,6 @@
 import tenseal as ts
 import pytest
+import numpy as np
```
philomath213, line 3: Since you moved `numpy` import here, I thing we should remove the import from other test functions, `test_vec_plain_matrix_mul` function still have import `numpy` there (line 375).

tests/tensors/test_ckks_vector.py
```diff
@@ -393,15 +394,42 @@ def test_vec_plain_matrix_mul(context, scale, vec, matrix):
...
+            ],
+        ),
+    ],
+)
+def test_vec_plain_matrix_mul_depth2(context, scale, vec, matrix1, matrix2):
+    context.generate_galois_keys()
+    ct = ts.ckks_vector(context, scale, vec)
+    result = ct @ matrix1 @ matrix2
+    expected = (np.array(vec) @ np.array(matrix1) @ np.array(matrix2)).tolist()
+    assert _almost_equal(result.decrypt(), expected, 1), "Matrix multiplciation is incorrect."
```
philomath213, line 430:
```suggestion
    assert _almost_equal(result.decrypt(), expected, 1), "Matrix multiplication is incorrect."
```

Issue #58, opened 2020-05-14
Provide a SEAL API binding to set scale manually after a multiplication


dhuynh95:
# Context

Currently after a multiplication has been done, there is no way to set the scale manually to allow further computation such as addition with a plaintext. 

Example : 
```python
# First do a multiplication and rescale
evaluator.multiply(ctx, ptx, output)
evaluator.rescale_to_next_inplace(output)
evaluator.mod_switch_to_inplace(ptx, output.parms_id())

# This is not allowed currently
output.scale(scale)

# Therefore this will fail because seal won't allow two objects of different scale to be summed
evaluator.add_plaintext_inplace(output, ptx)
```

# Proposed solution

I think a simple binding such as ctx.scale(scale) or ctx.set_scale(scale) would be enough to solve this problem.

bcebere:
Fixed with https://github.com/OpenMined/TenSEAL/pull/59

Added a `scale` property to Plaintext/Ciphertext classes, now you can do something like

`output.scale = ptx.scale`

PR #59, opened 2020-05-14
add support for setting the ct/pt scale

bcebere:
Fix for https://github.com/OpenMined/TenSEAL/issues/58


tenseal/sealapi/sealapi_encrypt.cpp
```diff
@@ -204,9 +204,7 @@ void bind_encrypt_decrypt(pybind11::module &m) {
         .def("parms_id", py::overload_cast<>(&Plaintext::parms_id, py::const_))
         .def("scale", py::overload_cast<>(&Plaintext::scale, py::const_))
         .def("set_scale",
-             [](Plaintext &obj, double scale) {
-                 obj.scale() = scale;
-             })
+             [](Plaintext &obj, double scale) { obj.scale() = scale; })
```
youben11, line 207: Can't we have something like property `.def_property("name", &Pet::getName, &Pet::setName)` from https://pybind11.readthedocs.io/en/stable/classes.html#instance-and-static-fields ?

bcebere: Nice, I didn't know about that

PR #60, opened 2020-05-16
Auto Relin, Rescale and ModSwitch

### LINKED ISSUE #38: Holds the scale in TenSEALContext
This issue is mainly related to the use of the CKKS scheme

### Motivation

1. API would be simpler if the user provides the `scale` once at context creation, this wasn't the case because we thought that the user would need to choose the scale for each tensor.
1. In a secure computation which is controlled by a single TenSEALContext, we will generally wants to have the same scale for all our ciphertexts.
1. After some multiplication and rescaling operations, we will sometimes need to assume that all ciphertexts are on the same scale, while this is not true, the ratio between the actual scales should be around 1 if the initial scales used for encoding were the same.

### Proposed solution

Hold the scale in the TenSEALContext for CKKS: this can also help us advise the user of choosing coeff modulus that are ~= sclae (e.g. scale=2**20 & coeff_mod_bit_sizes=[X, 20, 20, X]). Every multiplication should rescale the ciphertext using the next coeff modulus. If we assume that the initial scale is a power of two and that the coeff_modulus are of the same bit size as the initial scale, then all ciphertext should have a scale ~= init_scale no matter their multiplicative depth.

Ping: @philomath213 @s0l0ist

### LINKED ISSUE #49: Configure TenSEAL to auto relin/rescale
### Context

We want to both make relinearization/rescaling transparent to the user and offer experienced users flexibility to implement their ideas. 

### Possible solution

A good place to configure this would be the TenSEALContext object where you can for example call `context.[enable|disable]_auto_[relin|rescale]()` to turn on/off relin/rescaling.

See also #35

youben11:
After introducing the global_scale in #57, this PR allow for configuring the TenSEALContext to do automatically relin, rescale and modulus switching.

They are all switched on by default, they can be turned off by accessing the attributes `auto_relin`, `auto_rescale` and `auto_mod_switch` which will be globally shared across all ciphertexts (this can break at some point when doing parallel computation: a TODO was left)

The only changes to the API was the introduction of the previous cited attributes.

Fixes #38 
Fixes #49 
Related #35


tenseal/tensors/ckksvector.cpp
```diff
@@ -90,6 +90,11 @@ CKKSVector& CKKSVector::add_inplace(CKKSVector to_add) {
         throw invalid_argument("can't add vectors of different sizes");
     }
 
+    if (this->context->is_auto_mod_switch() &&
```
bcebere, line 93: Maybe the if condition can be separated in a dedicated function/method, something like `should_set_to_same_mod`.

tenseal/tensors/ckksvector.cpp
```diff
@@ -176,13 +198,23 @@ CKKSVector& CKKSVector::mul_inplace(CKKSVector to_mul) {
...
-    // relineraize after ciphertext-ciphertext multiplication
-    this->context->evaluator->relinearize_inplace(this->ciphertext,
-                                                  this->context->relin_keys());
+    if (this->context->is_auto_relin()) {
+        this->context->evaluator->relinearize_inplace(
+            this->ciphertext, this->context->relin_keys());
+    }
+
+    if (this->context->is_auto_rescale()) {
+        this->context->evaluator->rescale_to_next_inplace(this->ciphertext);
```
bcebere, line 215: Shouldn't we check if the ciphertext got to last_parms_id()? It will throw an exception when it gets there.

Is there a test to capture the situation?

tenseal/tensors/ckksvector.cpp
```diff
@@ -176,13 +198,23 @@ CKKSVector& CKKSVector::mul_inplace(CKKSVector to_mul) {
...
+        this->ciphertext.scale() = this->init_scale;
```
bcebere, line 216: Should we make init_scale `const double`?

Keep in mind, this won't work for consts in constructors
`A::A(double scale){
   this->init_scale = scale;
}`

Instead, you have to leverage the initializer lists mechanism.
``A::A(double scale) : init_scale(scale){}``

tenseal/tensors/ckksvector.cpp
```diff
@@ -204,9 +236,20 @@ CKKSVector& CKKSVector::mul_plain_inplace(vector<double> to_mul) {
...
+    if (this->context->is_auto_mod_switch() &&
+        this->ciphertext.parms_id() != plaintext.parms_id()) {
+        set_to_same_mod(this->context, this->ciphertext, plaintext);
+    }
+
     this->context->evaluator->multiply_plain_inplace(this->ciphertext,
                                                      plaintext);
 
+    if (this->context->is_auto_rescale()) {
+        this->context->evaluator->rescale_to_next_inplace(this->ciphertext);
```
bcebere, line 249: same as above:
 - should we move the condition to a separate function?
 - should we check for last_parms_id?

tenseal/utils/utils.h
```diff
@@ -0,0 +1,52 @@
...
+
+/*
+Apply modulus switching to the ciphertext (or plaintext) having the higher
+modulus.
+*/
+template <typename T>
+void set_to_same_mod(shared_ptr<TenSEALContext> context, Ciphertext& ct,
+                     T& other) {
+    size_t idx1 =
+        context->seal_context()->get_context_data(ct.parms_id())->chain_index();
```
bcebere, line 36: get_context_data returns nullptr if it doesn't find the parms_id. Should we check for that to prevent a crash?

tenseal/utils/utils.h
```diff
@@ -0,0 +1,52 @@
...
+    size_t idx2 = context->seal_context()
```
bcebere, line 37: Might be easier with a lambda function. Something like

```
auto get_chain_index = [&] (const auto& obj) -> std::optional<size_t>{
  auto ctx_data = context->seal_context()->get_context_data(obj.parms_id());
  if (ctx_data == nullptr) {
    return {}; 
  }
  return ctx_data->chain_index();
};

size_t idx1 = get_chain_index(ct).value_or(/*other value ?*/);
size_t idx2 = get_chain_index(other).value_or(/*other value ?*/);
```


(sorry, I haven't tried to compile, but it should work - it's just a suggestion).
Here, I've added a lambda function that returns an optional chain_index.
``` [&] ``` - means it captures the current context, you can access the local variables.
```optional<size_t>``` - means it can hold a size_t value, otherwise it won't use the memory.

File: tenseal/utils/utils.h [hunk 7]
```diff
@@ -0,0 +1,52 @@
...
+                      ->get_context_data(other.parms_id())
+                      ->chain_index();
+
+    if (idx1 == idx2) return;
```
bcebere, line 41: maybe we should use something like ct_idx and other_idx.
Because SEAL throws an exception if you mix the mod_switch_to_inplace parameters, just to be clear.

tenseal/tensealcontext.h
```diff
@@ -214,6 +225,13 @@ class TenSEALContext {
     */
     double _scale = -1;
 
+    /*
+    Switches for automatic relinearization, rescaling, and modulus switching
+    */
```
bcebere, line 230: Maybe in the future we can use a bitmask for keeping the internal flags, for momery improvements.

something like
```
enum {
  flag_auto_relin = 1 << 0,
  flag_auto_rescale = 1 << 1,
  flag_auto_mod_switch = 1 << 2,
}
 uint8_t _flags = flag_auto_mod_switch | flag_auto_rescale | flag_auto_relin;```

tenseal/tensors/ckksvector.cpp
youben11: we leave it to SEAL to throw an error when needed, but yeah, there is no test for this, I will add one

tenseal/tensors/ckksvector.cpp
youben11: I think it's more explicit to leave it like this, isn't it? what's the advantage of separating it?

tenseal/tensors/ckksvector.cpp
youben11: but the constructor doesn't just put the scale value there, there is some checks before.

tenseal/tensors/ckksvector.cpp
bcebere: Just because it's a similar condition in multiple places, that's all

tenseal/utils/utils.h
youben11: while we should check for this, I'm not quite sure what should be returned...

tenseal/tensealcontext.h
youben11: Yeah better, I will work on this

tenseal/tensors/ckksvector.cpp
youben11: I added an inline function (thought it may be better inline), what do you think?

tenseal/utils/utils.h
youben11: the question is what value should we return here, couldn't figure it out yet

bcebere: since we are relying on SEAL exceptions anyway, we could throw an exception here too. 
For example, instead of returning {} in the lambda function. And optional is not needed there, in this scenario.

tenseal/tensors/ckksvector.cpp
bcebere: Yes, I think it is easier to follow.

Maybe the check can be moved to `set_to_same_mod` completely.

youben11: `set_to_same_mod` should do its job unconditionally of `auto_mod_switch` (as used in utils/matrix_ops.h) and therefore should be kept outside

PR #65, opened 2020-05-17
move all the encoder logic to the tensealencoder class

### LINKED ISSUE #50: TenSEALEncoder improvements
We can further improve the encoder interface by:

- providing encode/decode templates to the contexts.
- simplifying the tensor/context logic by forwarding the encode/decode calls directly to the Encoder factory.(instead of getting an encoder object first)

bcebere:
## Description

Thank you for your contribution to the TenSEAL repository.
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

If your pull request closes a GitHub issue, then set its number below.

Fixes #50

## Type of change

Please mark options that are relevant.

- [ ] Added/Modified tutorials
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Refactoring
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

* [ ] I have added tests for my changes
* [x] I did follow the [contribution guidelines](#)

### DIFF
--- a/tenseal/tensealencoder.h
+++ b/tenseal/tensealencoder.h
@@ -39,7 +39,11 @@ class TenSEALEncoder {
     template <class CKKSEncoder>
     void encode(std::vector<double>& vec, Plaintext& pt,
                 std::optional<double> optscale = {}) {
-        auto scale = optscale.value_or(global_scale());
+        double scale = 1.0;
+        if (optscale.has_value())
+            scale = optscale.value();
+        else
+            scale = global_scale();
 
         auto encoder = this->get<CKKSEncoder>();
         encoder->encode(vec, scale, pt);
--- a/tenseal/tensealencoder.h
+++ b/tenseal/tensealencoder.h
@@ -72,7 +72,7 @@ class TenSEALEncoder {
     // Example: if using coeff_mod_bit_size of [60,40,40,60],
     // the global scale should be set for 2**40
     void global_scale(double scale) {
-        if (scale < 1.0) throw std::invalid_argument("invalid scale value");
+        if (scale < 0) throw std::invalid_argument("invalid scale value");
 
         this->_scale = scale;
     }
--- a/tenseal/tensealencoder.h
+++ b/tenseal/tensealencoder.h
@@ -37,8 +37,10 @@ class TenSEALEncoder {
     }
 
     template <class CKKSEncoder>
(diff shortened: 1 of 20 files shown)


tenseal/tensealencoder.h
```diff
@@ -25,6 +26,41 @@ class TenSEALEncoder {
...
+    */
+    template <class T>
+    void encode(std::vector<int64_t>& vec, Plaintext& pt) {
+        auto encoder = this->get<T>();
+        encoder->encode(vec, pt);
+    }
+
+    template <class CKKSEncoder>
+    void encode(std::vector<double>& vec, Plaintext& pt, double scale = 0) {
+        if (scale == 0) scale = pow(2, 40);  // TODO: get from TenSEALContext
```
youben11, line 41: TenSEALEncoder can be created using TenSEALContext instead of SEALContext so that we can fetch global_scale here

tenseal/tensealencoder.h
```diff
@@ -25,6 +26,65 @@ class TenSEALEncoder {
...
+
+    // throws invalid_argument exception if the scale is incompatible with
+    // current primes used. Example: if using coeff_mod_bit_size of
+    // [60,40,40,60], the global scale should be set for 2**40
+    void global_scale(double scale) {
+        auto parms_id = _context->first_parms_id();
+        auto ctx_data = _context->get_context_data(parms_id);
+        if (static_cast<int>(log2(scale)) >=
+            ctx_data->total_coeff_modulus_bit_count())
+            throw std::invalid_argument("invalid scale value");
```
youben11, line 78: I don't think we should do anything about it at this point, errors should be forwarded to SEAL

youben11: we can maybe force it to be positive here...

bcebere: ok. I was following the commented TODO.

SEAL throws an exception anyway, from a similar condition.

youben11: Ah, the TODO was about checking if the scale matches the coeff[1], as you may know, rescaling is dividing by each prime, and if we use primes with bit_size different than the scale, then we will probably end with a bad decryption. We shouldn't catch error that are checked by SEAL but rather focus on specific ones for TenSEAL.

Don't bother yourself with the TODO, a full strategy for checking those value need to be designed first, there is a lot to think of, and we can defer this to a later stage.

PR #76, opened 2020-06-17
Dot product for CKKSVector

### LINKED ISSUE #74: A aggregation function for an encrypted vector (add_many/multiply_many)
## Feature Description
A single encrypted TenSEAL vector should be able to be naively aggregated by either a sum or product. 

## Is your feature request related to a problem?
This would be especially useful for common operations such as dot product and euclidean distance. 

## What alternatives have you considered?
Currently the only known method I understand might for summation work is to perform a matrix multiplication with an identity vector which is not ideal.

## Additional Context
Ideally the operation would look something like 
`encrypted_vec.add_many()`

youben11:
## Description

Adding dot product operations for CKKSVector, we can dot product with another CKKSVector (encrypted) or a simple List (plain), there is both methods to return a new result or run the operations in place. Below is an example usage, refer to the added tests for more about how to use it.

```python
first_vec = tenseal.ckks_vector(context, [1, 2, 3, 4])
# second_vec can also be a simple python list for plain operations
second_vec = tenseal.ckks_vector(context, [4, 3, 2, 1])

# return a new result
result = first_vec.dot(second_vec)
# do the operation in place
first_vec.dot_(second_vec)

# expected result after decryption is [20]
```

@Christopher-Wang the feature and API are all set, however, I still need to implement a more efficient algorithm.

Fixes #74 

## Checklist


Christopher-Wang:
This is a great solution as many thinks could be reformulated as a dot product. However, would it also be possible to expose the sum_vector function on its own as well?

youben11:
Yeah, we can easily expose that as well. I should expose that and have the efficient algorithm working during the weekend hopefully.

Christopher-Wang:
Perfect. I started a PR to include that functionality and am looking to merge that into this branch. Unfortunately I forgot to tag it so until

PR #77, opened 2020-06-18
Add summation aggregation with dot product

Christopher-Wang:
## Description
Exposed the sum_vector operation to the python api. This is so that a more efficient summation can be performed (rather than perform a dot product on a plain identity vector). This introduces a new class of operations of CKKS vectors that only operate on one operand and return a vector of size one (aggregations). In the future other aggregations could be implemented (such as multiplications for a geometric mean).

## Affected Dependencies
No dependencies are required to change.

## How has this been tested?
I built the Dockerfile-py36 and ran the tests in interactive mode.

## Checklist

### DIFF
--- a/tenseal/binding.cpp
+++ b/tenseal/binding.cpp
@@ -100,6 +100,8 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
         .def("dot", &CKKSVector::dot_product_plain)
         .def("dot_", &CKKSVector::dot_product_inplace)
         .def("dot_", &CKKSVector::dot_product_plain_inplace)
+        .def("sum", &CKKSVector::sum)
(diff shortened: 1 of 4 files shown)


tianhan4:
This seems only to work when the vector size is less than slot_count/2, guaranteeing at least one repeated list followed.

youben11:
The summation algorithm doesn't require to have the vector repeated in the ciphertext. I just ran a test locally and it seems to work pretty fine, can you please open a new issue with the code that's getting a false result?

tianhan4:
I think you are right. I've made a mistake. Thought all values in the slots for the original vector should be the summation. In fact, the size has shrunk to 1. Sorry for your trouble.

Issue #46, opened 2020-05-11
Throw error when encrypting empty vectors


youben11:
Current implementation allows for encrypting an empty vector, which actually doesn't make sense. We should block the user when trying to do so.

IanQS:
Would you prefer an exception from within C++ or within Python?

youben11:
C++ please :). All the library features should be implemented in C++, Python is just a binding and can have specific features that only apply for it

PR #92, opened 2020-07-07
Different size vector operations

youben11:
## Description
We sometimes need to compute over encrypted vectors of different size,  this isn't defined when the vectors have sizes `n1` and `n2` which are both greater than 1, however, a vector of size 1 can be added for example with a vector of size `n` by replicating the first slot `n` times. This PR introduces this feature by adding a replication function in the encrypted domain, and perform the operations as usual after that.

It's worth noting that this operation of replication is quite costly, it requires a multiplication and `log(n)` rotation and addition.

ping @apignet

## How has this been tested?
- Added tests

## Checklist


tenseal/tensors/ckksvector.cpp
```diff
@@ -85,7 +85,15 @@ CKKSVector& CKKSVector::add_inplace(CKKSVector to_add) {
     }
 
     if (this->size() != to_add.size()) {
```
bcebere, line 87: can to_add have length 0?

tenseal/tensors/ckksvector.cpp
```diff
@@ -153,7 +161,15 @@ CKKSVector& CKKSVector::sub_inplace(CKKSVector to_sub) {
     }
 
     if (this->size() != to_sub.size()) {
-        throw invalid_argument("can't sub vectors of different sizes");
+        if (this->size() == 1) {
+            this->replicate_first_slot_inplace(to_sub.size());
+            return this->sub_inplace(to_sub);
```
bcebere, line 166: why not just continue instead of adding one more call?

tenseal/tensors/ckksvector.cpp
```diff
@@ -85,7 +85,15 @@ CKKSVector& CKKSVector::add_inplace(CKKSVector to_add) {
     }
 
     if (this->size() != to_add.size()) {
-        throw invalid_argument("can't add vectors of different sizes");
+        if (this->size() == 1) {
+            this->replicate_first_slot_inplace(to_add.size());
+            return this->add_inplace(to_add);
+        } else if (to_add.size() == 1) {
+            CKKSVector replicated = to_add.replicate_first_slot(this->size());
+            return this->add_inplace(replicated);
```
bcebere, line 93: why not just continue instead of making one more call?

tenseal/tensors/ckksvector.cpp
```diff
@@ -221,7 +237,15 @@ CKKSVector& CKKSVector::mul_inplace(CKKSVector to_mul) {
     }
 
     if (this->size() != to_mul.size()) {
-        throw invalid_argument("can't multiply vectors of different sizes");
+        if (this->size() == 1) {
+            this->replicate_first_slot_inplace(to_mul.size());
+            return this->mul_inplace(to_mul);
```
bcebere, line 242: same about the return

tenseal/tensors/ckksvector.cpp
```diff
@@ -352,4 +376,32 @@ CKKSVector& CKKSVector::matmul_plain_inplace(
...
+    CKKSVector new_vector = *this;
+    return new_vector.replicate_first_slot_inplace(n);
+}
+
+CKKSVector& CKKSVector::replicate_first_slot_inplace(size_t n) {
+    // mask
+    vector<double> mask = {1};
+    // TODO: remove this after resolving issue of transparent ciphertext
+    // which adds a 1 at the end
+    for (size_t i = 0; i < n - 1; i++) mask.push_back(0);
```
bcebere, line 389: Why not
vector<double> mask(n, 0);
mask[0] = 1;
?

youben11: True!

tenseal/tensors/ckksvector.cpp
youben11: Actually yeah, thanks for spotting this! We should open an issues about this, as all operations will need this additional check and maybe we will refactor all those check in some way.

philomath213: currently, it's possible to have an empty vector.
check #46

tenseal/tensors/ckksvector.cpp
philomath213: if you are talking about the 2nd check, the multiplication could be either Nx1 or 1xN, so we need also to check if the 2nd vector is of size 1.

tenseal/tensors/ckksvector.cpp
youben11: Yeah there is already an issue about encrypted vectors, but we should also check plain vectors, which can be passed like `encrypted_vector + []`

youben11: Although it doesn't work, but maybe a better error message can be used here, and since the checks are repeated many time, might be worth some refactoring

tenseal/tensors/ckksvector.cpp
```diff
@@ -85,7 +86,13 @@ CKKSVector& CKKSVector::add_inplace(CKKSVector to_add) {
     }
 
     if (this->size() != to_add.size()) {
-        throw invalid_argument("can't add vectors of different sizes");
+        if (this->size() == 1) {
```
bcebere, line 89: this check is identical in all methods, maybe it can be separated in a private method.

youben11: yeah, that's what I was referring to in my comments :point_up: , we should later refactor all the checks into a function with a common error message like "can't operate on vectors of different sizes");

PR #94, opened 2020-07-09
Right arithmetic operators for CKKS vector

philomath213:
## Description
Add python binding for right arithmetic operators (e.g. __radd__, __rsub__) in order to be able to do arithmetic operation where the plaintext is the right operand.
Since `Subtraction` operation is anti-commutative, I implemented an additional method for CKKS vector `rsub_plain` because
right subtraction operator need to negate the vector then do an addition with left operand.

### DIFF
--- a/tenseal/binding.cpp
+++ b/tenseal/binding.cpp
@@ -42,11 +42,11 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
         py::arg("encryption_parameters"));
 
     py::class_<BFVVector>(m, "BFVVector")
-        .def(py::init<shared_ptr<TenSEALContext>&, vector<int64_t>>())
+        .def(py::init<shared_ptr<TenSEALContext> &, vector<int64_t>>())
         .def("size", &BFVVector::size)
         .def("save_size", &BFVVector::save_size)
         .def("decrypt", py::overload_cast<>(&BFVVector::decrypt))
-        .def("decrypt", py::overload_cast<const std::shared_ptr<SecretKey>&>(
+        .def("decrypt", py::overload_cast<const std::shared_ptr<SecretKey> &>(
                             &BFVVector::decrypt))
         .def("add", &BFVVector::add)
         .def("add_", &BFVVector::add_inplace)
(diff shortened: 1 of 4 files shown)


tenseal/binding.cpp
```diff
@@ -141,6 +144,10 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
         .def("__sub__", py::overload_cast<double>(&CKKSVector::sub_plain))
         .def("__sub__",
              py::overload_cast<vector<double>>(&CKKSVector::sub_plain))
+        // subtraction operation is anticommutative
+        .def("__rsub__", py::overload_cast<double>(&CKKSVector::rsub_plain))
```
youben11, line 148: I think we can define them here instead of defining an rsub method in CKKSVector

```suggestion
        .def("__rsub__", [](double left_operand) {
                 CKKSVector new_vector = this->negate();
                 new_vector.add_plain_inplace(left_operand);
                 return new_vector;
             })
```

tenseal/binding.cpp
```diff
@@ -141,6 +144,10 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
...
+        .def("__rsub__",
```
youben11, line 149: same for this

tenseal/tensors/ckksvector.cpp
```diff
@@ -206,6 +206,20 @@ CKKSVector CKKSVector::sub_plain(double to_sub) {
...
+    return new_vector;
+}
+
+CKKSVector CKKSVector::rsub_plain(double left_operand) {
+    CKKSVector new_vector = this->negate();
+    new_vector.add_plain_inplace(left_operand);
+
+    return new_vector;
+}
+
```
youben11, line 222: Can be removed after the definition in the bindings

tenseal/tensors/ckksvector.h
```diff
@@ -95,6 +95,13 @@ class CKKSVector {
     CKKSVector sum();
     CKKSVector& sum_inplace();
 
+    /*
+    Since subtraction operation is anticommutative, right subtraction operator
+    need to negate the vector then do an addition with left operand.
+    */
+    CKKSVector rsub_plain(double to_sub);
+    CKKSVector rsub_plain(vector<double> to_sub);
+
```
youben11, line 104: Can be removed after the definition in the bindings

tenseal/tensors/ckksvector.cpp
```diff
@@ -206,6 +206,20 @@ CKKSVector CKKSVector::sub_plain(double to_sub) {
     return new_vector;
 }
 
+CKKSVector CKKSVector::rsub_plain(vector<double> left_operand) {
```
bcebere, line 209: this create an unnecessary copy of the vector.

youben11: but we can't do an inplace operation, or how can we prevent the copy?

philomath213: `this->negate()` will return a new vector, and the following `add_plain_inplace` won't create a new vector.
we have to create a new vector, because we can't modify the vector when it's on the right side of the operation.
example: 
v  = [1,2,3] # ckks encrypted vector
x = 2 - v  # scaler operation 
x == [1, 0, -1]
in this case v shouldn't change.

bcebere: I was referring that left_operand should be const vector<double>&, now it is copied to the parameter.

tenseal/binding.cpp
philomath213: I thought about that before, but since we need to access `this` pointer we can do that in a lambda function.

youben11: True! This should work:

```suggestion
        .def("__rsub__", [](CKKSVector vec, double left_operand) {
                 // vec should be a copy so it might be safe to do inplace
                 vec->negate_inplace();
                 vec.add_plain_inplace(left_operand);
                 return vec;
             })
```

tenseal/tensors/ckksvector.cpp
philomath213: yeah, you are write, we need to fix that in all functions

Issue #91, opened 2020-07-06
Optimizing use of immutable vectors


youben11:
## Description
As pointed out by @bcebere in #86 , we are passing some immutable vector by values instead of const reference which can be more efficient.

PR #96, opened 2020-07-10
use replicate_first_slot instead of replicate_first_slot_inplace

philomath213:
## Description
PR #92 introduced "different size vector operations", these operation were modify the 2nd operand by calling `replicate_first_slot_inplace` instead of `replicate_first_slot` in the case where the 2nd operand is of size 1.
This bug didn't have an effect because we are coping the argument instead of passing them as const reference (see #91).


youben11:
The operand is already copied into the function, so the outside variable isn't modified in the inplace operation, and this should work without issues, however, using the non-inplace operation shouldn't work, so that's problematic...

philomath213:
The tests didn't fail because of using `replicate_vector` during encrypting CKKS vector.

PR #97, opened 2020-07-10
Passing immutable vectors by reference

### LINKED ISSUE #91: Optimizing use of immutable vectors

philomath213:
## Description
Passing immutable vectors by reference `(const &)` instead of passing them by value.

Fixes #91 .

### DIFF
--- a/tenseal/tensors/bfvvector.cpp
+++ b/tenseal/tensors/bfvvector.cpp
@@ -84,14 +84,14 @@ BFVVector& BFVVector::add_inplace(BFVVector to_add) {
     return *this;
 }
 
-BFVVector BFVVector::add_plain(vector<int64_t> to_add) {
+BFVVector BFVVector::add_plain(const vector<int64_t>& to_add) {
     BFVVector new_vector = *this;
     new_vector.add_plain_inplace(to_add);
 
     return new_vector;
 }
 
-BFVVector& BFVVector::add_plain_inplace(vector<int64_t> to_add) {
+BFVVector& BFVVector::add_plain_inplace(const vector<int64_t>& to_add) {
     if (this->size() != to_add.size()) {
         throw invalid_argument("can't add vectors of different sizes");
     }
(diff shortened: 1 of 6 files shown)


tenseal/tensors/bfvvector.cpp
```diff
@@ -178,23 +178,24 @@ BFVVector& BFVVector::mul_inplace(BFVVector to_mul) {
...
 
     Plaintext plaintext;
+    vector<int64_t> new_vec_to_mul(to_mul);
     // prevent transparent ciphertext by adding a non-zero value
-    if (to_mul.size() + 1 <= this->context->slot_count<BatchEncoder>())
-        to_mul.push_back(1);
-    this->context->encode<BatchEncoder>(to_mul, plaintext);
+    if (new_vec_to_mul.size() + 1 <= this->context->slot_count<BatchEncoder>())
+        new_vec_to_mul.push_back(1);
+    this->context->encode<BatchEncoder>(new_vec_to_mul, plaintext);
```
youben11, line 198: Can you add a TODO here to mention that we won't need to create a new vector when we remove the transparent hack so that we don't forget about it

tenseal/tensors/ckksvector.cpp
```diff
@@ -291,23 +291,24 @@ CKKSVector CKKSVector::mul_plain(double to_mul) {
...
+CKKSVector& CKKSVector::mul_plain_inplace(const vector<double>& to_mul) {
     if (this->size() != to_mul.size()) {
         throw invalid_argument("can't multiply vectors of different sizes");
     }
     // prevent transparent ciphertext by adding a non-zero value
-    if (to_mul.size() + 1 <= this->context->slot_count<CKKSEncoder>())
-        to_mul.push_back(1);
+    vector<double> new_vec_to_mul(to_mul);
+    if (new_vec_to_mul.size() + 1 <= this->context->slot_count<CKKSEncoder>())
+        new_vec_to_mul.push_back(1);
```
youben11, line 301: same here

bcebere:
LGTM!

tenseal/tensors/bfvvector.cpp
philomath213: done

PR #98, opened 2020-07-12
Catch transparent ct and replace with encrypted zero

### LINKED ISSUE #36: Should we disable SEAL_THROW_ON_TRANSPARENT_CIPHERTEXT?
Default build of SEAL enforce SEAL_THROW_ON_TRANSPARENT_CIPHERTEXT to on which will throw an error whenever the result ciphertext of an evaluation is transparent.

Current implementation workaround issues regarding transparent ciphertexts in vector operations #33. An open question is whether we should disable it completely? Does our use cases get affected?

youben11:
## Description

When multiplying with a plain vector of all zeros, an exception is raised about the resulting ciphertext which is transparent. As discussed in #36, we now catch this exception and replace the resulting ct with an encryption of zero.

Fixes #36 

## How has this been tested?
- Automated tests

## Checklist

### DIFF
--- a/tenseal/tensors/bfvvector.cpp
+++ b/tenseal/tensors/bfvvector.cpp
@@ -190,16 +190,16 @@ BFVVector& BFVVector::mul_plain_inplace(const vector<int64_t>& to_mul) {
         throw invalid_argument("can't multiply vectors of different sizes");
     }
 
-    // TODO: rmeove this after fixing #36
-    // prevent transparent ciphertext by adding a non-zero value
     Plaintext plaintext;
-    vector<int64_t> new_vec_to_mul(to_mul);
-    if (new_vec_to_mul.size() + 1 <= this->context->slot_count<BatchEncoder>())
-        new_vec_to_mul.push_back(1);
-    this->context->encode<BatchEncoder>(new_vec_to_mul, plaintext);
+    this->context->encode<BatchEncoder>(to_mul, plaintext);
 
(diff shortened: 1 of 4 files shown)


tenseal/tensors/bfvvector.cpp
```diff
@@ -190,16 +190,16 @@ BFVVector& BFVVector::mul_plain_inplace(const vector<int64_t>& to_mul) {
...
-    // prevent transparent ciphertext by adding a non-zero value
     Plaintext plaintext;
-    vector<int64_t> new_vec_to_mul(to_mul);
-    if (new_vec_to_mul.size() + 1 <= this->context->slot_count<BatchEncoder>())
-        new_vec_to_mul.push_back(1);
-    this->context->encode<BatchEncoder>(new_vec_to_mul, plaintext);
-
-    this->context->evaluator->multiply_plain_inplace(this->ciphertext,
-                                                     plaintext);
+    this->context->encode<BatchEncoder>(to_mul, plaintext);
```
bcebere, line 194: SEAL evaluator's docs specify the transparent ct exception for more operations, like negate, add, relinearize etc. 

should we cover all?

youben11: I'm still not sure what are the use case for those, so far using CKKS, multiplication has been the most important operation to deal with, so I would love to discover those failing use cases first before catching them :smile:

PR #99, opened 2020-07-13
CKKSVector Polynomial Evaluation

### LINKED ISSUE #95: Polynomial evaluation for CKKSVector
## Feature Description
Add a method for evaluating univariate polynomials where the variable is a CKKSVector. The method should meet the following criteria:
- Use square and multiply for exponentiation for both efficiency and minimizing the multiplicative depth
- The method should receive a list of coefficients of variable size (n+1), with the first being the constant term, and the last being the coefficient of x^n
- coefficients 1 and 0 shouldn't involve a multiplication

For the naming, `polyval` seems like a good name (borrowed from the numpy API)

youben11:
## Description
This PR introduce a new method for CKKSVector to be able to evaluate a polynomial over all the vector values (`encrypted_vector.polyval(list_of_polynomial_coefficients)`)

Fixes #95 

## How has this been tested?
- Added tests

## Checklist


tenseal/tensors/ckksvector.cpp
```diff
@@ -412,4 +415,60 @@ CKKSVector& CKKSVector::replicate_first_slot_inplace(size_t n) {
...
+    }
+
+    // set result accumulator to the constant coefficient
+    vector<double> cst_coeff(this->size(), coefficients[0]);
+    CKKSVector result(this->context, cst_coeff, this->init_scale);
+
+    // pre-compute squares of x
+    CKKSVector x = *this;
+    int max_square = static_cast<int>(floor(log2(degree)));
+    cout << "max square " << max_square << endl;
```
bcebere, line 453: Do we need stdout logging here?

tenseal/tensors/ckksvector.cpp
```diff
@@ -412,4 +415,60 @@ CKKSVector& CKKSVector::replicate_first_slot_inplace(size_t n) {
...
+    vector<CKKSVector> x_squares;
+    x_squares.reserve(max_square + 1);
```
bcebere, line 455: Why not directly resize? and access the vector directly using [] instead of push_back operations

tenseal/tensors/ckksvector.cpp
```diff
@@ -412,4 +415,60 @@ CKKSVector& CKKSVector::replicate_first_slot_inplace(size_t n) {
...
+    // we can multiply by 0, or return the encryption of zero
+    if (degree == -1) {
+        // we set the vector to the encryption of zero
+        this->context->encryptor->encrypt_zero(this->ciphertext);
+        this->ciphertext.scale() = this->init_scale;
+        return *this;
+    }
+
+    // set result accumulator to the constant coefficient
+    vector<double> cst_coeff(this->size(), coefficients[0]);
```
bcebere, line 447: this creates a vector of size this->size() with **all** the values set to coefficients[0]. is that the expected behavior?

tenseal/tensors/ckksvector.cpp
youben11: leftovers...

tenseal/tensors/ckksvector.cpp
youben11: yep, to encrypt it later and have operations done element-wise

tenseal/tensors/ckksvector.cpp
```diff
@@ -412,4 +415,60 @@ CKKSVector& CKKSVector::replicate_first_slot_inplace(size_t n) {
...
+    return new_vector.polyval_inplace(coefficients);
+}
+
+CKKSVector& CKKSVector::polyval_inplace(const vector<double>& coefficients) {
+    if (coefficients.size() == 0) {
+        throw invalid_argument(
+            "the coefficients vector need to have at least one element");
+    }
+
+    int degree = coefficients.size() - 1;
```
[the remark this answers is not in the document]
youben11: How should we do that?

tenseal/tensors/ckksvector.cpp
youben11: True!

tenseal/tensors/ckksvector.cpp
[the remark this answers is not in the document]
bcebere: My confusion here is that coefficients.size() returns size_t, which is an unsigned int64.

Assigning to int should raise some warnings from any compiler, I think.

And you can use 
```
int degree = static_cast<int>(coefficients.size()) - 1;
```

to prevent the warnings.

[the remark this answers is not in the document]
youben11: Ah yeah of course, didn't saw that I didn't cast, thank you!

tenseal/tensors/ckksvector.cpp
```diff
@@ -412,4 +415,58 @@ CKKSVector& CKKSVector::replicate_first_slot_inplace(size_t n) {
...
+    }
+
+    // set result accumulator to the constant coefficient
+    vector<double> cst_coeff(this->size(), coefficients[0]);
+    CKKSVector result(this->context, cst_coeff, this->init_scale);
+
+    // pre-compute squares of x
+    CKKSVector x = *this;
+    int max_square = static_cast<int>(floor(log2(degree)));
+    vector<CKKSVector> x_squares(max_square + 1);
```
bcebere, line 453: I think the build fails here now, because it tries to call CKKSVector with the default constructor.
If 
```
 vector<CKKSVector> x_squares;
x_squares.resize(max_square + 1);
```

doesn't work, ignore this

tenseal/tensors/ckksvector.cpp
youben11: **Update:** So it's a little complicated for this, reserve is more convenient for our usage, as resize need to init the objects in the vector, and we doesn't want this, however, reserve will only allocate the necessary memory and push_back won't have to allocate memory dynamically.

bcebere: Yeah, my observation made more sense with vector of pointers.
Sorry about that.

youben11: Thank you!

PR #100, opened 2020-07-13
CKKS square and power functions

philomath213:
## Description
Implement square and power functions for CKKS vector with minimal multiplication depth.
Now it's possible to compute the power of a CKKS vector using `.power` method or `**` operator.
The power function compute recursively the result by splitting the computation into two balanced tree, as a result we will have a result ciphertext with minimal multiplication depth.
There is a special case when the power is zero, in this case a new CKKS vector of ones is returned, while keeping the same size.

## Checklist


tenseal/tensors/ckksvector.cpp
```diff
@@ -84,6 +84,68 @@ CKKSVector& CKKSVector::negate_inplace() {
...
+
+    if (this->context->auto_rescale()) {
+        this->context->evaluator->rescale_to_next_inplace(this->ciphertext);
+        this->ciphertext.scale() = this->init_scale;
+    }
+
+    return *this;
+}
+
+CKKSVector CKKSVector::power(int power) {
```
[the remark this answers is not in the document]
philomath213: yeah, you are right, `unsigned int` will remove the overhead of checking the negative values of power.

tenseal/tensors/ckksvector.cpp
```diff
@@ -84,6 +84,68 @@ CKKSVector& CKKSVector::negate_inplace() {
...
+
+    if (power == 2) {
+        this->square_inplace();
+        return *this;
+    }
+
+    int closest_power_of_2 = 1 << static_cast<int>(floor(log2(power)));
+    power -= closest_power_of_2;
+    if (power == 0) {
+        this->power_inplace(closest_power_of_2 / 2)
```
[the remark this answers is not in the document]
philomath213: I thought it will require one more multiplication depth, but it won't.
in this case `this->power_inplace(closest_power_of_2 / 2).square_inplace()` would be better.

tenseal/tensors/ckksvector.cpp
```diff
@@ -84,6 +84,68 @@ CKKSVector& CKKSVector::negate_inplace() {
...
+            this->ciphertext, *this->context->relin_keys());
+    }
+
+    if (this->context->auto_rescale()) {
+        this->context->evaluator->rescale_to_next_inplace(this->ciphertext);
+        this->ciphertext.scale() = this->init_scale;
+    }
+
+    return *this;
+}
```
youben11, line 108: why not calling mul_inplace which should take care of relin/rescale?

tenseal/tensors/ckksvector.cpp
```diff
@@ -84,6 +84,68 @@ CKKSVector& CKKSVector::negate_inplace() {
...
+
+    int closest_power_of_2 = 1 << static_cast<int>(floor(log2(power)));
+    power -= closest_power_of_2;
+    if (power == 0) {
+        this->power_inplace(closest_power_of_2 / 2)
+            .mul_inplace(this->power(closest_power_of_2 / 2));
+    } else {
+        this->power_inplace(power).mul_inplace(this->power(closest_power_of_2));
+    }
+    return *this;
```
youben11, line 146: We can use the compute_polynomial_term from utils, we can refactor the name, but calling it with the power, coeffcient as 1, and the precomputed squares of x should compute the power with the most efficient circuit possible.

tests/tensors/test_ckks_vector.py
```diff
@@ -43,6 +43,98 @@ def test_negate_inplace(context, plain_vec):
...
+    [
+        ([], 2),
+        ([0], 3),
+        ([0, 1, -1, 2, -2], 0),
+        ([1, -1, 2, -2], 1),
+        ([1, -1, 2, -2], 2),
+        ([1, -1, 2, -2], 3),
+        ([1, -2, 3, -4], 1),
+        ([1, -2, 3, -4], 2),
+        ([1, -2, 3, -4], 3),
```
youben11, line 57: raising to the power of 4 should be also possible with the default parameters, we need a test-case for this to make sure the circuit is optimum. Same for the other tests

philomath213: Yeah, this should work with power of 4 as well.

tenseal/tensors/ckksvector.cpp
philomath213: calling mul_inplace requires coping the ciphertext (i.e. `this->mul_inplace(*this)`).

tenseal/tensors/ckksvector.cpp
philomath213: `compute_polynomial_term` is particular case of `power` which have extra overhead.
I think keeping `power` function as it is would be better for both efficiency and readability.

tenseal/binding.cpp
```diff
@@ -87,6 +87,12 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
         .def("neg", &CKKSVector::negate)
         .def("neg_", &CKKSVector::negate_inplace)
         .def("neg_inplace", &CKKSVector::negate_inplace)
+        .def("square", &CKKSVector::square)
+        .def("square_", &CKKSVector::square_inplace)
+        .def("square_inplace", &CKKSVector::square_inplace)
+        .def("pow", &CKKSVector::power)
+        .def("pow_", &CKKSVector::power_inplace)
+        .def("pow_inplace", &CKKSVector::power_inplace)
```
youben11, line 95: The python API have have this inplace operations terminated by a `_` (e.g. `add_`) only and we aren't providing `function_name_inplace` kind of methods.

tenseal/tensors/ckksvector.cpp
```diff
@@ -84,6 +84,63 @@ CKKSVector& CKKSVector::negate_inplace() {
...
+    new_vector.power_inplace(power);
+
+    return new_vector;
+}
+
+CKKSVector& CKKSVector::power_inplace(unsigned int power) {
+    // if the power is zero, return a new encrypted vector of ones
+    if (power == 0) {
+        vector<double> ones(this->size(), 1);
+        *this = CKKSVector(this->context, ones, this->init_scale);
```
youben11, line 121: I think you can here just replace the ciphertext (`ciphertext = this->encrypt(...)`) instead of creating a whole new CKKSVector.

tenseal/tensors/ckksvector.cpp
```diff
@@ -84,6 +84,63 @@ CKKSVector& CKKSVector::negate_inplace() {
...
+        this->square_inplace();
+        return *this;
+    }
+
+    int closest_power_of_2 = 1 << static_cast<int>(floor(log2(power)));
+    power -= closest_power_of_2;
+    if (power == 0) {
+        this->power_inplace(closest_power_of_2 / 2).square_inplace();
+    } else {
+        this->power_inplace(power).mul_inplace(this->power(closest_power_of_2));
```
youben11, line 139: this is a little bit confusing how it's evaluated, maybe something like this would be more readable?

```suggestion
        CKKSVector closest_pow2_vector = this->power(closest_power_of_2);
        this->power_inplace(power).mul_inplace(closest_pow2_vector);
```

tests/tensors/test_ckks_vector.py
youben11: still missing a testcase of 4 here

Issue #72, opened 2020-06-01
Serialise, deserialise; for pickle, deepcopy


DreamingRaven:
## Feature Description
I feel it would be vital to be able to deep-copy objects like context, private key, and ciphertext.
Similarly It will also be vital to real use to be able to pickle, and unpickle, context, private key, and ciphertext, or at the least save to a file-like object, but not necessarily a file itself.

## Is your feature request related to a problem?
I would like to save directly to a non local system such as a database the necessary objects to encrypt, evaluate, and decrypt ciphertext, without having to write to the local filesystem, thus I need to create a file-like object to store elsewhere. Similarly to this end this will involve serialisation and desirialisation, probably from pickle, which currentley does not work with the pybin11 bindings here. This also prevents things like python deep-copying any of the tenseal objects which is necessary under certain use cases like several workers copying from the same object to evaluate/ compute some function.

## What alternatives have you considered?
Saving to a file and loading that file to a database; time consuming, IO intensive, bottlenecked, and not easily scalable.

## Additional Context
Here is a unit test showcasing the current inability to pickle, and deep-copy, unless I am misunderstanding how it is to be done here:
``test.py``
```
import copy
import unittest


class seal_tests(unittest.TestCase):
    """Unit test class aggregating all tests for the seal class."""

    def get_context_bfv(self):
        import tenseal as ts
        context = ts.context(ts.SCHEME_TYPE.BFV, poly_modulus_degree=4096,
                             plain_modulus=1032193)
        return context

    def test_copy_context_bfv(self):
        context = self.get_context_bfv()
        copy.deepcopy(context)

    def get_keys_bfv(self):
        context = self.get_context_bfv()
        sk = context.secret_key()
        return (context, sk)

    def test_copy_keys_bfv(self):
        context, sk = self.get_keys_bfv()
        copy.deepcopy(sk)

    def get_ciphertext(self):
        import tenseal as ts
        context = self.get_context_bfv()
        plain_vector = [60, 66, 73, 81, 90]
        encrypted_vector = ts.bfv_vector(context, plain_vector)
        return encrypted_vector

    def test_copy_ciphertext(self):
        ciphertext = self.get_ciphertext()
        copy.deepcopy(ciphertext)


if __name__ == "__main__":
    # run all the unit-tests
    unittest.main()
```
when run ``python3 ./test.py``
```
EEE
======================================================================
ERROR: test_copy_ciphertext (__main__.seal_tests)
----------------------------------------------------------------------
[...]

PR #101, opened 2020-07-14
TensealContext serialization

bcebere:
## Description

- [x] Protobuf support
- [x] Context serialization/deserialization
- [x] gtests
- [x] Python support
- [x] Tests
- [x] Bonus: make_public_context crash. Fixed with a scoped guard.

Related to https://github.com/OpenMined/TenSEAL/issues/72

## How has this been tested?
 - gtest
 - pytest

## Checklist


tenseal/proto/tensealcontext.proto
```diff
@@ -0,0 +1,38 @@
...
+
+
+//TenSEAL Private parameters
+message TenSEALPrivateProto {
+    // Serialized secret key
+    bytes secret_key = 1;
+    // Relin keys flag
+    bool relin_keys = 2;
+    // Galois keys flag
+    bool galois_keys = 3;
```
youben11, line 13: the name is somehow misleading here, it seems like we are talking about the keys themselves, where it's whether the keys are generated or not, can we rename to key_generated or something similar?

tenseal/utils/scope.h
```diff
@@ -0,0 +1,35 @@
...
+    void dismiss() noexcept { f = nullptr; }
+
+    scope_guard(const scope_guard&) = delete;
+    void operator=(const scope_guard&) = delete;
+
+   private:
+    std::function<void()> f;
+};
+}  // namespace tenseal
+#endif
```
youben11, line 35: Can I know what this is doing?

tests/test_serialization.py
```diff
@@ -0,0 +1,136 @@
...
+    assert context.auto_relin == True
+    assert context.auto_rescale == True
+    assert context.auto_mod_switch == False
+
+    orig_context.auto_mod_switch = True
+    context = duplicate(orig_context)
+
+    assert context.auto_relin == True
+    assert context.auto_rescale == True
+    assert context.auto_mod_switch == True
```
youben11, line 136: Can we add tests to test the functioning of the keys after serialization, like duplicating a private context with galois and relin keys (which should regenerate them after deserialization) then doing dot product operation (which should involve both relin and galois keys) and see if the result is correct.

tenseal/utils/scope.h
bcebere: Scoped guards implement a classic C++ idiom - RAII, Resource Acquisition is Initialization. shared_ptr is another example of that.
Long story short, It will get a lambda function on the constructor, and it will call it on its destructor.
The use case is simple: I want to add some logic on a function return. For example, I want to free some resources, a create a scope guard with a lambda that frees the resources, and no matter how many returns I have in the function, the resources will get freed on exit, because the variable's destructor gets called.


My issue here was that, while writing serialization test cases, I observed that when calling make_context_public with a true value for galois or relin keys, there was a crash

This is the reason, I am to blame
https://github.com/OpenMined/TenSEAL/blame/master/tenseal/tensealcontext.cpp#L124

I was making the secret key null before creating the galois/relin keys.


The fix is to create a scoped guard that will make the secret key null on exit, no matter how many conditional returns there will be. Something like this, and I'm sure it will make the secret key null everytime
https://github.com/OpenMined/TenSEAL/blob/protobuffs/tenseal/tensealcontext.cpp#L166

I hope the explanation makes sense.

youben11: Thanks for the explanation. It's fine to keep it in this PR, but feel free to open another one if it's convenient for you.

tenseal/binding.cpp
```diff
@@ -217,10 +218,8 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
         )",
              py::arg("poly_modulus_degree"), py::arg("plain_modulus"),
              py::arg("coeff_mod_bit_sizes") = vector<int>())
-        .def("load", py::overload_cast<const char *>(&TenSEALContext::Create))
-        .def("save_public", &TenSEALContext::save_public, "save public keys.")
-        .def("save_private", &TenSEALContext::save_private,
-             "save private keys.")
+        .def("new",
```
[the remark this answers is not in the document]
bcebere: I renamed
save -> serialize
new -> deserialize, and made it static.

tenseal/binding.cpp
```diff
@@ -235,16 +234,38 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
...
+             })
+        .def(py::pickle(
+            [](const std::shared_ptr<TenSEALContext> &p) {
+                return py::make_tuple(p->save());
+            },
+            [](py::tuple t) {
+                if (t.size() != 1) throw std::runtime_error("Invalid state!");
+                auto p = TenSEALContext::Create(t[0].cast<std::string>());
+                return p;
+            }));
```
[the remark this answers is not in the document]
bcebere: I change copy/deepcopy to just recreate a new object from the raw protobuffer, instead of the serialized version.

.github/workflows/tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt update
-        sudo apt install cmake clang
+        sudo apt install cmake clang libprotobuf-dev protobuf-compiler
```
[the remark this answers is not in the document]
bcebere: updated README

.github/workflows/tests.yml
```diff
@@ -52,6 +52,7 @@ jobs:
         pytest -v --ignore=tests/sealapi ./tests
     - name: Test with gtest
       run: |
+        cd tenseal/proto && cmake . && make && cd ../../
```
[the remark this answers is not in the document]
bcebere: Unfortunately, here I'm missing something about cmake builds. For some reason, it doesn't generate the proto headers for the build otherwise.

On the other hand, setup.py generates them just fine, and I cannot spot the actual difference.

tests/test_serialization.py
```diff
@@ -0,0 +1,189 @@
...
+    orig_context = ctx()
+    context = duplicate(orig_context)
+
+    assert context.is_private() is True, "TenSEALContext should be private"
+    assert context.public_key() is not None, "TenSEALContext shouldn't be None"
+
+    orig_context.make_context_public(generate_galois_keys=False, generate_relin_keys=False)
+    context = duplicate(orig_context)
+    assert context.is_public() is True, "TenSEALContext should be public"
+
```
philomath213, line 46:
```suggestion

    orig_context = ctx()
    context = duplicate(orig_context)
    context.make_context_public(generate_galois_keys=False, generate_relin_keys=False)
    assert orig_context.is_private() is True, "original TenSEALContext should stay private"
```

philomath213: additional check here would be great, in order to verify that the context object is copied properly.

tests/test_serialization.py
```diff
@@ -0,0 +1,189 @@
...
+    context = duplicate(orig_context)
+    assert isinstance(context.galois_keys(), ts.GaloisKeys), "Galois keys should be set"
+
+    orig_public_context = ctx()
+    orig_public_context.make_context_public(generate_galois_keys=True, generate_relin_keys=False)
+    context = duplicate(orig_public_context)
+
+    assert isinstance(context.galois_keys(), ts.GaloisKeys), "Galois keys should be set"
+
+
```
philomath213, line 62:
```suggestion
    orig_context = ctx()
    context = duplicate(orig_context)
    context.generate_galois_keys()
    with pytest.raises(ValueError) as exc_info:
        orig_context.galois_keys()
    assert exc_info.type is ValueError, "orginal context shouldn't have Galois keys"

```

philomath213: same here

tests/test_serialization.py
```diff
@@ -0,0 +1,189 @@
...
+    orig_public_context = ctx()
+    orig_public_context.make_context_public(generate_galois_keys=False, generate_relin_keys=True)
+    context = duplicate(orig_public_context)
+    assert isinstance(context.relin_keys(), ts.RelinKeys), "Relin keys should be set"
+
+    orig_public_context = ctx()
+    orig_public_context.make_context_public(generate_galois_keys=True, generate_relin_keys=True)
+    context = duplicate(orig_public_context)
+    assert isinstance(context.relin_keys(), ts.RelinKeys), "Relin keys should be set"
+    assert isinstance(context.galois_keys(), ts.GaloisKeys), "Galois keys should be set"
```
philomath213, line 80:
```suggestion
@pytest.mark.parametrize("duplicate", [deep_copy, simple_copy, internal_copy, recreate, pickled,])
def test_generate_relin_keys(duplicate):
    # by default, Relin keys will be generated on context creation
    orig_context = ctx()
    context = duplicate(orig_context)
    assert isinstance(context.relin_keys(), ts.RelinKeys), "Relin keys should be set"

    orig_public_context = ctx()
    orig_public_context.make_context_public(generate_galois_keys=False, generate_relin_keys=False)
    context = duplicate(orig_public_context)
    with pytest.raises(ValueError) as exc_info:
        context.relin_keys()
    assert exc_info.type is ValueError, "context shouldn't have Relin keys"

    orig_public_context = ctx()
    orig_public_context.make_context_public(generate_galois_keys=False, generate_relin_keys=True)
    context = duplicate(orig_public_context)
    assert isinstance(context.relin_keys(), ts.RelinKeys), "Relin keys should be set"
```

philomath213: by default, Relin keys will be generated on context creation, so no need to call `generate_relin_keys` method.

tests/test_serialization.py
```diff
@@ -0,0 +1,189 @@
...
+
```
philomath213, line 91:
```suggestion


@pytest.mark.parametrize("duplicate", [deep_copy, simple_copy, internal_copy, recreate, pickled,])
def test_generate_galois_relin_keys(duplicate):
    orig_public_context = ctx()
    orig_public_context.make_context_public(generate_galois_keys=True, generate_relin_keys=True)
    context = duplicate(orig_public_context)
    assert isinstance(context.relin_keys(), ts.RelinKeys), "Relin keys should be set"
    assert isinstance(context.galois_keys(), ts.GaloisKeys), "Galois keys should be set"

    orig_context = ctx()
    context = duplicate(orig_context)
    context.make_context_public(generate_galois_keys=True, generate_relin_keys=True)
    with pytest.raises(ValueError) as exc_info:
        orig_context.relin_keys()
        orig_context.galois_keys()
    assert exc_info.type is ValueError, "orginal context shouldn't have Relin and Galois keys"

```

philomath213: having a separate test for both Galois and Relin keys would be great.

tenseal/serialization.h
```diff
@@ -0,0 +1,56 @@
...
+ * Compatible SEAL types: Ciphertext, Plaintext, SecretKey, PublicKey,
+ *GaloisKeys, RelinKeys.
+ **/
+template <class T>
+std::string SEALSerialize(const T& sealobj) {
+    std::stringstream stream;
+    sealobj.save(stream);
+
+    return stream.str();
+}
```
philomath213, line 21: I think `EncryptionParameters` object is also compatible with this function, isn't it?

.github/workflows/tests.yml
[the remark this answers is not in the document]
philomath213: when I build the tenseal without `BUILD_TEST` argument, the build is successful.
example:
`cmake .`
`make`

we don't have to build `tenseal_proto` manually, but when passing the argument `BUILD_TEST=TRUE`, the build is failed, and we have to build `tenseal_proto` manually.

I think we are missing something in **Unit Tests** section in CMakeLists.txt file.

PR #105, opened 2020-07-19
Upgrade SEAL to 3.5.6

xueyumusic:
## Description
The versions of SEAL from 3.5.1 to 3.5.6 has [fixed many bugs and some optimizations](https://github.com/microsoft/SEAL/blob/master/CHANGES.md). This PR tried to upgrade SEAL from 3.5.1 to 3.5.6.
Then updated places are:
1. Since SEAL 3.5.6 removed `util/polyarith.h` and `util/polyarithmod.h`, removing related parts.
2. Updated some places related with `util/polyarithsmallmod.h` changes.
3. Updated some function call params using `ConstRNSIter` and `RNSIter` because of seal function param changes
4. Add `MultiplyUIntModOperand` python binding and update `coeff_div_plain_modulus` return type

## Affected Dependencies
SEAL lib

## How has this been tested?
pytest -v --ignore=tests/sealapi ./tests
cmake . -D BUILD_TEST=TRUE
make && CTEST_OUTPUT_ON_FAILURE=1 make test
pytest -v ./tests/sealapi

test cases under `tests` dir

## Checklist
- [Y ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [Y] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [ ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [Y] My changes are covered by tests

### DIFF
--- a/tenseal/sealapi/sealapi_util_namespace.cpp
+++ b/tenseal/sealapi/sealapi_util_namespace.cpp
@@ -8,10 +8,9 @@
 #include <seal/util/croots.h>
 #include <seal/util/hash.h>
 #include <seal/util/hestdparms.h>
+#include <seal/util/iterator.h>
 #include <seal/util/numth.h>
 #include <seal/util/pointer.h>
-#include <seal/util/polyarith.h>
-#include <seal/util/polyarithmod.h>
 #include <seal/util/polyarithsmallmod.h>
 #include <seal/util/rlwe.h>
 #include <seal/util/scalingvariant.h>
@@ -139,7 +138,7 @@ void bind_util_namespace(pybind11::module &m) {
              [](const BaseConverter &obj, const std::vector<std::uint64_t> &in,
                 std::size_t count) {
                  std::vector<std::uint64_t> out(obj.obase_size() * count);
-                 obj.fast_convert_array(in.data(), count, out.data(),
+                 obj.fast_convert_array(ConstRNSIter(in.data(), count), RNSIter(out.data(), count),
                                         MemoryManager::GetPool());
 
                  return out;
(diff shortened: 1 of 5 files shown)


tenseal/sealapi/sealapi_util_namespace.cpp
```diff
@@ -159,33 +160,42 @@ void bind_util_namespace(pybind11::module &m) {
...
+                std::size_t count, std::uint64_t *destination) {
+                 obj.fastbconv_sk(ConstRNSIter(input, count),
+                                  RNSIter(destination, count),
+                                  MemoryManager::GetPool());
              })
         .def("sm_mrq",
-             [](const RNSTool &obj, const std::uint64_t *input, std::size_t count,
-                std::uint64_t *destination) {
-                 obj.sm_mrq(ConstRNSIter(input, count), RNSIter(destination, count), MemoryManager::GetPool());
+             [](const RNSTool &obj, const std::uint64_t *input,
```
bcebere, line 174: I think pybind doesn't map uint64_t* to a vector, it will consider it a single value.

I think it better to use vector instead of (uint64, size_t). and call it using (input.data(), input.size()).


LE: Now I see that the issue was there all along, it's not new. It was documented here https://github.com/OpenMined/TenSEAL/blob/master/tenseal/sealapi/sealapi_util_namespace.cpp#L30

but we didn't fix it everywhere.

Please replace the (std::uint64_t *, size_t) parameters with std::vector, to have the full functionality

tenseal/sealapi/sealapi_util_namespace.cpp
```diff
@@ -716,6 +505,12 @@ void bind_util_namespace(pybind11::module &m) {
      * } "util/hash.h"
      *******************/
 
+    py::class_<MultiplyUIntModOperand>(m, "MultiplyUIntModOperand",
```
bcebere, line 508: can we have some sanity (py)tests for this class?

tenseal/sealapi/sealapi_util_namespace.cpp
xueyumusic: Thans for review. I updated the codes to use std::vector as params, also since the seal's `ConstRNSIter` and `RNSIter` need a poly_modulus_degree param which is not equal with vector size, so add `poly_modulus_degree` in some places.

tenseal/sealapi/sealapi_util_namespace.cpp
xueyumusic: Add a `test_util_multiplyuintmodoperand_sanity` test case in test_util.py

tenseal/sealapi/sealapi_util_namespace.cpp
```diff
@@ -152,45 +153,65 @@ void bind_util_namespace(pybind11::module &m) {
...
         .def("fastbconv_sk",
-             [](const RNSTool &obj, const std::uint64_t *input,
-                std::uint64_t *destination) {
-                 obj.fastbconv_sk(input, destination, MemoryManager::GetPool());
+             [](const RNSTool &obj, const std::vector<uint64_t> &input,
+                std::vector<uint64_t> &destination,
+                std::size_t poly_modulus_degree) {
+                 obj.fastbconv_sk(
+                     ConstRNSIter(input.data(), poly_modulus_degree),
+                     RNSIter(destination.data(), poly_modulus_degree),
```
bcebere, line 176: Shouldn't destination vector be resized to input.size()?

now, by default, it will be empty. isn't that going to lead to a crash?

Issue #88, opened 2020-07-06
Automate tests on MacOS and Windows


youben11:
## Feature Description
We need to test on all platforms (Linux, MacOS, Windows) on every PR.

## Is your feature request related to a problem?
The build on MacOS and Windows was only ran on release, which made us spot problems on release time, instead of catching them early on every codebase change.

ramesht007:
Hey, @youben11 can I work on this issue

youben11:
Hi @ramesht007! Thank you! Feel free to DM me on slack if you need further information :)

PR #108, opened 2020-07-21
tests on MacOS and Windows #88

ramesht007:
## Description
created automated tests for Windows and mac for issue #88


.github/workflows/tests.yml
```diff
@@ -98,3 +98,177 @@ jobs:
...
+        python -m pip install --upgrade pip
+        pip install -r requirements_dev.txt
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 --config=.flake8 --exclude=tenseal/sealapi,tests/sealapi .
+        black --check --exclude "sealapi" tenseal/ tests/
+        # exit-zero treats all errors as warnings.
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=100 --statistics
+    - name: Lint with clang-format
```
bcebere, line 126: I don't think we need to duplicate the linting logic to all platforms, Linux is enough. It would fail anyway

.github/workflows/tests.yml
```diff
@@ -98,3 +98,177 @@ jobs:
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements_dev.txt
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 --config=.flake8 --exclude=tenseal/sealapi,tests/sealapi .
+        black --check --exclude "sealapi" tenseal/ tests/
+        # exit-zero treats all errors as warnings.
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=100 --statistics
```
bcebere, line 212: same for MacOS

.github/workflows/tests.yml
```diff
@@ -98,3 +98,177 @@ jobs:
     - name: Test with pytest
       run: |
         pytest -v ./tests/sealapi
+  
+  TenSEAL_WIN:
+
+    runs-on: windows-latest
```
bcebere, line 104: I think there a lot of duplicate code here, you can you something global like

```    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [windows-latest, macos-latest, ubuntu-18.04]
```

And keep a single TenSEAL

You can see a similar example here https://github.com/OpenMined/PSI/blob/master/.github/workflows/tests.yml

.github/workflows/tests.yml
ramesht007: ok i will make changes

.github/workflows/tests.yml
ramesht007: same here

.github/workflows/tests.yml
ramesht007: ok I will make changes

.github/workflows/tests.yml
```diff
@@ -99,53 +100,6 @@ jobs:
...
-      
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
-      with:
-        python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
```
ramesht007: hey @bcebere I think these commands will give an error, can I remove them

.github/workflows/tests.yml
```diff
@@ -9,10 +9,11 @@ on:
 jobs:
   TenSEAL:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8]
+        os: [windows-latest, macos-latest, ubuntu-latest]
```
youben11, line 16: I don't think a matrix of OSes will work as each platform may differ in the setup phase

.github/workflows/tests.yml
```diff
@@ -22,8 +23,6 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        sudo apt update
-        sudo apt install cmake clang libprotobuf-dev protobuf-compiler
```
youben11, line 26: Linux requires installing some packages

.github/workflows/tests.yml
```diff
@@ -98,3 +97,57 @@ jobs:
...
+        pip install -r requirements_dev.txt
+    - name: Build the library
+      run: |
+        # fetch third_party libraries
+        git submodule init && git submodule update
+        # we need to install tenseal in order to build the cpp library
+        pip install .                
+    - name: Test with pytest
+      run: |
+        pytest -v ./tests/sealapi
```
youben11, line 126: Don't expect to setup Windows the same way as Linux, checkout this https://github.com/OpenMined/TenSEAL/blob/master/.github/workflows/pythonpublish-windows.yml

.github/workflows/tests.yml
ramesht007: so I should add each OS separately

.github/workflows/tests.yml
ramesht007: That I will add.... just removed it...so it should work on windows

.github/workflows/tests.yml
ramesht007: ok

.github/workflows/tests.yml
youben11: I would prefer having conditioning on Windows and doing separate build for it, but if that's not possible, removing the matrix might be necessary

ramesht007: I can add separate build for windows and mac both.... if it's ok

bcebere: You can add dedicated scripts for each matrix.os.

Something like

```
        run: .github/workflows/scripts/lint_cpp.sh
        if: ${{ matrix.os == 'ubuntu-18.04' }}
```
You can see an example here https://github.com/OpenMined/PSI/blob/2a43e74c07a1eb010006cd5b6621a9435c53a7a9/.github/workflows/tests.yml

.github/workflows/tests.yml
```diff
@@ -9,38 +9,26 @@ on:
...
-    - name: Lint with clang-format
-      run: |
-        # lint files (all .cpp and .h files) inplace
-        find ./tenseal/  \( -iname *.h -o -iname *.cpp \) -a -not -path "*sealapi*" | xargs clang-format -i -style='file'
-        # print changes
-        git diff
-        # already well formated if 'git diff' doesn't output
-        ! ( git diff |  grep -q ^ ) || return 1
+    - name: ubuntu-build
+      run: chmod +x .github/workflows/scripts/run_build_ubuntu.sh
```
bcebere, line 25: why chmod +x ? you can commit the flag

.github/workflows/scripts/run_build_ubuntu.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -e
+
+sudo apt update
+sudo apt install cmake clang libprotobuf-dev protobuf-compiler
+python -m pip install --upgrade pip
+pip install -r requirements_dev.txt
```
bcebere, line 8: you can build the library here too/run the tests

.github/workflows/tests.yml
ramesht007: without it's saying permission denied

.github/workflows/scripts/run_build_ubuntu.sh
ramesht007: ok

.github/workflows/tests.yml
bcebere: you can run chmod +x on your machine and commit the flag

ramesht007: ok

.github/workflows/scripts/run_build_ubuntu.sh
ramesht007: can you give an example, like how can I segregate the commands in the same file

.github/workflows/tests.yml
```diff
@@ -9,38 +9,24 @@ on:
...
-    - name: Lint with clang-format
-      run: |
-        # lint files (all .cpp and .h files) inplace
-        find ./tenseal/  \( -iname *.h -o -iname *.cpp \) -a -not -path "*sealapi*" | xargs clang-format -i -style='file'
-        # print changes
-        git diff
-        # already well formated if 'git diff' doesn't output
-        ! ( git diff |  grep -q ^ ) || return 1
+    - name: Install dependencies Ubuntu
+      run: chmod +x .github/workflows/scripts/depends_ubuntu.sh
```
bcebere, line 25: This doesn't install the depends, only sets the executable flag. You have to run this on you machine and remove chmod here

ramesht007: ohhhh ok

PR #109, opened 2020-07-22
add serialization support in tensors

bcebere:
## Description
Protobuff support for tensors

- [x] Support in BFVVector and CKKSVector
- [x] Python bindings
- [x] gtest
- [x] pytest

fixes https://github.com/OpenMined/TenSEAL/issues/72

## Checklist


tenseal/tensealcontext.cpp
```diff
@@ -243,6 +248,11 @@ bool TenSEALContext::auto_mod_switch() {
     return this->_auto_flags & flag_auto_mod_switch;
 }
 
+bool TenSEALContext::equals(const std::shared_ptr<TenSEALContext>& other) {
+    return this->parms() == other->parms() &&
+           this->is_public() == other->is_public();
+}
```
youben11, line 254: Maybe we can give this a deeper thought, what about the other attributes, the scale for example? And I don't understand why the contexts needs to be both the same type (public vs private)

tests/tensors/test_bfv_vector.py
```diff
@@ -1,12 +1,40 @@
 import tenseal as ts
```
youben11, line 1: Nice spot for the filename lol

tests/tensors/test_bfv_vector.py
```diff
@@ -320,18 +418,28 @@ def test_mul_plain(context, vec1, vec2):
...
 
 
-def test_size(context):
+@pytest.mark.parametrize(
+    "duplicate", [noop, deep_copy, simple_copy, internal_copy, recreate, pickled,]
+)
+def test_size(context, duplicate):
     for size in range(10):
         vec = ts.bfv_vector(context, [1] * size)
         assert vec.size() == size, "Size of encrypted vector is incorrect."
```
youben11, line 445: While I like the idea of integrating the serialization tests into the others, it's a lot to process for tests, I think adding separate test for the general use cases can cover the same functionalities while reducing tests time

tenseal/proto/tensors.proto
```diff
@@ -0,0 +1,25 @@
...
+    // The size of the encrypted vector
+    uint32 size = 2;
+    // The serialized ciphertext
+    bytes ciphertext = 3;
+};
+
+//CKKSVector parameters
+message CKKSVectorProto {
+    // The context protobuffer
+    TenSEALContextProto context = 1;
```
youben11, line 18: You are saving the context into the vectors, which can be avoided and can reduce a lot of space. An example use case is where someone wants to send a bunch of encrypted vectors for outsourced computation, so he will only need to send a single context, and multiple encrypted vectors, that when received should be linked to the received context.

Can we have a ts.ckks_vector_from(buffer, context) kind of API which links the vector to the specified context without the need to include it in the vectors' protobuff?

tenseal/tensealcontext.cpp
bcebere: It's enough to compare the public key and scale? I am not sure about the relin/galois/secret keys here.

tenseal/tensors/ckksvector.cpp
```diff
@@ -37,8 +39,19 @@ CKKSVector::CKKSVector(const CKKSVector& vec) {
     this->ciphertext = vec.ciphertext;
 }
 
-size_t CKKSVector::size() { return this->_size; }
-size_t CKKSVector::ciphertext_size() { return this->ciphertext.size(); }
+CKKSVector::CKKSVector(const string& ctx, const string& vec) {
+    this->load_context(ctx);
+    this->load(vec);
+}
```
youben11, line 45: The receiving machine will need to have multiple copies of the context in memory. It would be great if ctx is an actual TenSEALContext pointer object, that we only set without any deserialization, so that the receiving machine will deserialize the context, then link it to all the other vectors without replicating it.

I think a function `link_to_context` for vector might make sense here, to decouple the deserialization with this lonking logic, so maybe we deserialize normally, but the vector won't be operational until we link it to a context, does this make sense?

philomath213: I think an overload would be a better:
`CKKSVector::CKKSVector(shared_ptr<TenSEALContext> ctx, const string& vec)`
This will allow creating a vector using an exiting TenSEALContext as well as from a serialized TenSEALContext.

tenseal/tensors/ckksvector.cpp
```diff
@@ -150,7 +159,7 @@ CKKSVector CKKSVector::add(CKKSVector to_add) {
 }
 
 CKKSVector& CKKSVector::add_inplace(CKKSVector to_add) {
-    if (this->context != to_add.context) {
+    if (!this->context->equals(to_add.context)) {
```
philomath213, line 162: I think we don't have to check if the two context are equals since SEAL operations (add, sub, mult, negate) will check for context and the parameters internally.

tenseal/tensealcontext.cpp
```diff
@@ -240,6 +254,14 @@ bool TenSEALContext::auto_mod_switch() {
...
 }
 
+bool TenSEALContext::equals(const std::shared_ptr<TenSEALContext>& other) {
+    if (this->parms() != other->parms()) return false;
+
+    if (this->safe_global_scale() != other->safe_global_scale()) return false;
+
+    return true;
+}
+
```
philomath213, line 264: I'm not sure if this the right way to compare two TenSEALContext, since it's have a SEALContext inside.
SEAL will check internally the validation of context and encryption parameters while doing arithmetic operations as well as serialization.
I don't think we are going to compare two TenSEALContext at all, is there any use case for that?
ping @youben11

philomath213:
This is really great work, it will allow us to build and try client/server scenarios easily.
I just left few comments.
Good job @bcebere.

tenseal/tensors/ckksvector.cpp
youben11: We shouldn't create contexts per encrypted vector, context should be shared, so I think we shouldn't have a method for deserializing a context directly into a vector.

tenseal/tensealcontext.cpp
youben11: I'm actually not sure at all about that, as Bogdan pointed out in a previous comment, what about the other keys?

Also, if the scale isn't the same, operations aren't well defined, so we can't only leave it to SEAL to make the check. This needs more analysis actually and I would prefer to move it to another issue/PR to work on it specifically.

tenseal/tensors/__init__.py
```diff
@@ -22,6 +22,19 @@ def bfv_vector(context, plaintext_vector):
...
+    """Constructor method for the BFVVector object from a serialized protobuffer.
+    Args:
+        ctx_buff: The serialized context protobufer.
+        vec_buff: The serialized vector protobufer.
+
+    Returns:
+        BFVVector object.
+    """
+
+    return _ts_cpp.BFVVector(ctx_buff, vec_buff)
```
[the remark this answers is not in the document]
bcebere: @youben11  I've update the interface, please let me know if it's better this way.

tenseal/tensealcontext.cpp
bcebere: I'll revert it to the previous pointer comparison in this PR.

youben11: This will make it not work for our use case, I think you can leave it as is with equals method, but make equals always return True with a TODO comment

philomath213: Yeah, you are right
I forget about the scale, if it's different it will lead to wrong result.

bcebere: https://github.com/OpenMined/TenSEAL/issues/112

tenseal/tensors/ckksvector.cpp
bcebere: Added https://github.com/OpenMined/TenSEAL/issues/112 for a better review

tenseal/tensors/__init__.py
```diff
@@ -6,39 +6,79 @@
...
+        data: serialized protobuffer.
+
+    Returns:
+        BFVVector object.
+    """
+    if isinstance(context, _ts_cpp.TenSEALContext) and isinstance(data, bytes):
+        return _ts_cpp.BFVVector(context, data)
 
-    return _ts_cpp.BFVVector(context, plaintext_vector)
+    raise "Invalid BFV input types context: {} and vector: {}".format(type(context), type(data))
```
youben11, line 34: Sorry if I wasn't clear about this, I didn't meant we should have a single API functions, we still need both to differentiate them, but the user could have used both of them to do the same, as there was no python checking on the parameters, and Cpp overloading will be used either you pass a list or bytes, the method will return the object. However, we need different functions, `tensor` and `tensor_from` that take a list or bytes respectively, and can't take something else, the check should be done on the Python side and not in Cpp so that overloading isn't accidentally used in an undefined way.

tenseal/tensors/__init__.py
```diff
@@ -6,39 +6,91 @@
...
+        data: a list of integers to be encrypted.
 
     Returns:
         BFVVector object.
     """
+    if isinstance(context, _ts_cpp.TenSEALContext) and isinstance(data, list):
+        return _ts_cpp.BFVVector(context, data)
 
-    return _ts_cpp.BFVVector(context, plaintext_vector)
+    raise "Invalid BFV input types context: {} and vector: {}".format(type(context), type(data))
```
youben11, line 24:
```suggestion
    raise TypeError("Invalid BFV input types context: {} and vector: {}".format(type(context), type(data)))
```

tenseal/tensors/__init__.py
```diff
@@ -6,39 +6,91 @@
...
+        context: a TenSEALContext object, holding the encryption parameters and keys.
+        data: serialized protobuffer.
+
+    Returns:
+        BFVVector object.
+    """
+    if isinstance(context, _ts_cpp.TenSEALContext) and isinstance(data, bytes):
+        return _ts_cpp.BFVVector(context, data)
+
+    raise "Invalid BFV input types context: {} and vector: {}".format(type(context), type(data))
```
youben11, line 40:
```suggestion
    raise TypeError("Invalid BFV input types context: {} and vector: {}".format(type(context), type(data)))
```

tenseal/tensors/__init__.py
```diff
@@ -6,39 +6,91 @@
...
+            data: a list of floats to be encrypted.
+        Returns:
+            CKKSVector object.
+        """
+        if isinstance(context, _ts_cpp.TenSEALContext) and isinstance(data, list):
+            return _ts_cpp.CKKSVector(context, data)
+
+        raise "Invalid CKKS input types context: {} and vector: {}".format(
+            type(context), type(data)
+        )
```
youben11, line 61:
```suggestion
        raise TypeError("Invalid CKKS input types context: {} and vector: {}".format(
            type(context), type(data)
        ))
```

tenseal/tensors/__init__.py
```diff
@@ -6,39 +6,91 @@
...
+
+        Returns:
+            CKKSVector object.
+        """
+        if isinstance(context, _ts_cpp.TenSEALContext) and isinstance(data, list):
+            return _ts_cpp.CKKSVector(context, data, scale)
+
+        raise "Invalid CKKS(scale) input types context: {} and vector: {}".format(
+            type(context), type(data)
+        )
```
youben11, line 78:
```suggestion
        raise TypeError("Invalid CKKS(scale) input types context: {} and vector: {}".format(
            type(context), type(data)
        ))
```

tenseal/tensors/__init__.py
```diff
@@ -6,39 +6,91 @@
...
     Returns:
         CKKSVector object.
     """
-    if scale is None:
-        return _ts_cpp.CKKSVector(context, plaintext_vector)
-    return _ts_cpp.CKKSVector(context, plaintext_vector, scale)
+    if isinstance(context, _ts_cpp.TenSEALContext) and isinstance(data, bytes):
+        return _ts_cpp.CKKSVector(context, data)
+
+    raise "Invalid CKKS input types context: {} and vector: {}".format(type(context), type(data))
```
youben11, line 93:
```suggestion
    raise TypeError("Invalid CKKS input types context: {} and vector: {}".format(type(context), type(data)))
```

PR #111, opened 2020-07-22
Updated docker documentation, added .dockerignore

DreamingRaven:
## Description

TLDR: Did not change much, just changed documentation to run docker properly and added .dockerignore for good practice.

.dockerignore is always important to have similarly to .gitignore to
keep in this case the docker context free from unecessary bloat.

Updated the documentation to correct the docker build and run invocation
so that it keep the existing style of using a subdirectory to store the
dockerfiles but puts the context in the current parent directory so the
files can be copied in properly.

## Affected Dependencies
None

## How has this been tested?
- Describe the tests that you ran to verify your changes.
ran the new commands
```
sudo docker build -t tenseal . -f docker-images/Dockerfile-py38
```
and 
```
sudo docker run -it tenseal
```
- Provide instructions so we can reproduce.
please run the above two commands or:
```
sudo docker build -t tenseal . -f docker-images/Dockerfile-py38 && sudo docker run -it tenseal
```
- List any relevant details for your test configuration.

## Checklist
- [ ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [ ] My changes are covered by tests

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,1 @@
 tutorials
-third_party
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+tutorials
+third_party
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,6 @@
 tutorials
+build
+bin
+dist
+lib
+wheels
--- a/README.md
+++ b/README.md
@@ -101,7 +101,13 @@ $ docker container run --interactive --tty openmined/tenseal
 You can also build your custom image, this might be handy for developers working on the project
 
 ```bash
-$ docker image build --tag tenseal .
+$ docker build -t tenseal . -f docker-images/Dockerfile-py38
+```
+
+To interactively run this docker image as a container after it has been built you can run
+
+```bash
+$ docker run -it tenseal
 ```
 
 ## Support
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ $ docker container run --interactive --tty openmined/tenseal
 You can also build your custom image, this might be handy for developers working on the project
 
 ```bash
-$ docker build -t tenseal . -f docker-images/Dockerfile-py38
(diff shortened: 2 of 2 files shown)


DreamingRaven:
related to: https://github.com/OpenMined/TenSEAL/issues/72

.dockerignore
```diff
@@ -0,0 +1,2 @@
+tutorials
+third_party
```
youben11, line 2: I don't think those should be ignored, someone might want to run the docker image to test the tutorials which he expect to be there, third_party libraries also doesn't need to be downloaded if they are already on the host machine. Some example directories that might need to be ignored are build, bin, dist, lib, and wheels which might be generated in the working directory when installing and building the library.

DreamingRaven: Ah ok yeah, I misunderstood what third party was, I will remove that line.

As for tutorials, they are data, image, and jupyter notebook files from the looks of it. However they have no way to run jupyter  files in your docker containers, nor does the container expose a port for users to interact with the jupyter server if it was running, so they could play around with these tutorials even if the files were included. 
I will remove that line if you still want the file in, but I don't think its reasonable to expect an average user to be able to modify the docker container to be able to run jupyter notebooks since the container does not support it, and they are unnecessary for everyone else to my understanding.

youben11: True that we doesn't install juputer-notebook in the current images, maybe we should ignore them and build a specific image for that. Then we can leave it as is, but can you add [build, bin, dist, lib, wheels] also, those are the one I know about which are generated when I'm building and playing with the package, I don't want them to be included in the docker image.

DreamingRaven: Yeah no problem will add them now.

youben11: also, please merge master into your branch so that I can merge it back to master

DreamingRaven: done

youben11: Thank you!

Issue #106, opened 2020-07-19
Add Bazel support


bcebere:
## Feature Description
Bazel support would have numerous advantages in TenSEAL's build system.

## Pros
 - We would be extending the cmake build system, not replacing it.
 - Well-established ML tools like PyTorch or Tensorflow support Bazel too, and it would make the integration seamless there.
 - The integration with SEAL would be easy, no submodule required, as in this example. Bazel has bindings to other build systems as well https://github.com/OpenMined/PIR/blob/master/third_party/seal.BUILD
 - It would be a major improvement for integration/deployment in Bazel system, it would install all the depends without any hassle, just the TenSEAL reference in their bazel file, as in this example https://github.com/OpenMined/PIR
- It supports multiple languages, like C++/Python/Go/Rust and so on.
 - It supports pytest/gtest, no need for installing/running extra tools. Just `bazel test` and it executes all the tests in the repository, no matter the language/framework.

## Cons
 - pip support exists, but I am not sure it works fine with pybind, I wasn't able to make it work in this example https://github.com/OpenMined/PSI/blob/py-setuptools/setup.py
 - Bazel can be unstable, the API might change between updates.

bcebere:
I'll work on this one.

PR #116, opened 2020-07-26
Bazel build

bcebere:
## Description
Extend the build support to Bazel

TODO:

- [x] Build library and bindings.
- [x] Build  tests.
- [x]  Add Bazel tests to workflow.
- [x] Reorganize the c++ library, without breaking the existing CMake or changing the API.
- [x] MacOS tests work without any additional code.  https://github.com/OpenMined/TenSEAL/blob/bazel/.github/workflows/tests.yml#L102

fixes https://github.com/OpenMined/TenSEAL/issues/106

## Checklist


requirements_dev.txt
```diff
@@ -6,3 +6,5 @@ pytest>=5.4.1
 pytest-benchmark
 pytest-benchmark[histogram]
 setuptools>=41.0.0
+pygal
+pygaljs
```
youben11, line 10: where is this required?

tenseal/__init__.py
```diff
@@ -1,7 +1,11 @@
 """TenSEAL is a library for doing homomorphic encryption operation on tensors.
 """
 
-import _tenseal_cpp as _ts_cpp
+try:
+    import _tenseal_cpp as _ts_cpp
+except ImportError:
+    import tenseal._tenseal_cpp as _ts_cpp
+
```
youben11, line 8: So bazel put it inside the python package, actually this would be even cool to have without bazel, I just didn't know how to do it with setuptools

tenseal/binding.cpp
```diff
@@ -1,14 +1,13 @@
...
-#include <seal/seal.h>
 
 #include <memory>
 #include <vector>
 
-#include "tenseal/tensealcontext.h"
-#include "tenseal/tensors/bfvvector.h"
-#include "tenseal/tensors/ckksvector.h"
-#include "tenseal/utils/utils.h"
+#include "seal/seal.h"
```
youben11, line 7: is this okay to include it like this? shouldn't it be considered like an installed library?

tenseal/cpp/context/tensealcontext.h
```diff
@@ -1,11 +1,10 @@
 #ifndef TENSEAL_TENSEALCONTEXT_H
 #define TENSEAL_TENSEALCONTEXT_H
```
youben11, line 2: Maybe rename to TENSEAL_CONTEXT_TENSLEACONTEXT_H

tenseal/cpp/tensors/utils/utils.cpp
```diff
@@ -1,47 +1,13 @@
-#include <seal/seal.h>
-
 #include <memory>
-#include <thread>
```
youben11, line 4: this one is used in get_concurrency

tenseal/sealapi/sealapi_helpers.cpp
```diff
@@ -121,7 +121,6 @@ void bind_helpers(pybind11::module &m) {
      * "seal/serialization.h" {
      ***/
     py::enum_<compr_mode_type>(m, "COMPR_MODE_TYPE", py::module_local())
-        .value("DEFLATE", compr_mode_type::deflate)
```
youben11, line 124: Is this because we disabled ZLIB?

third_party/seal.BUILD
```diff
@@ -0,0 +1,29 @@
...
+    visibility = ["//visibility:public"]
+)
+
+cmake_external(
+   name = "seal",
+   cmake_options = [
+        "-DSEAL_USE_CXX17=17",
+        "-DSEAL_USE_INTRIN=ON",
+        "-DSEAL_USE_MSGSL=OFF",
+        "-DSEAL_USE_ZLIB=OFF",
```
youben11, line 15: Are we sure we want to disable ZLIB? I've seen some updates to the tests regarding this, while the Cmake build is using ZLIB, we should still be testing this, so how about enabling it here and keep the tests? I'm actually unaware of advantages of disabling it so would love to know more about the different cases

requirements_dev.txt
bcebere: pytest-benchmark seems to require them, I use requirements_dev.txt
 for bazel depends too.

tenseal/binding.cpp
bcebere: If I include it with <seal/seal.h>, it will search the system headers too.
If I include it with "seal/seal.h", it will search just the linked headers in the build process.

It can lead no weird bugs due to system artifacts, we should use our headers.

tests/python/sealapi/main.py
```diff
@@ -0,0 +1,6 @@
+import sys
+import os
+import pytest
+
+if __name__ == "__main__":
+    sys.exit(pytest.main(["-s", "-v", "-x", os.path.dirname(os.path.abspath(__file__))]))
```
youben11, line 6: Maybe comment what this is doing and specify that it's needed specifically by Bazel

tests/python/tenseal/main.py
```diff
@@ -0,0 +1,6 @@
+import sys
+import os
+import pytest
+
+if __name__ == "__main__":
+    sys.exit(pytest.main(["-s", "-v", "-x", os.path.dirname(os.path.abspath(__file__))]))
```
youben11, line 6: Maybe comment what this is doing and specify that it's needed specifically by Bazel

CMakeLists.txt
```diff
@@ -27,24 +27,27 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR})
...
 set(SOURCES
-    ${TENSEAL_BASEDIR}/tensealcontext.cpp
-    ${TENSEAL_BASEDIR}/tensors/bfvvector.cpp
-    ${TENSEAL_BASEDIR}/tensors/ckksvector.cpp
-    ${TENSEAL_BASEDIR}/utils/utils.cpp
+    ${TENSEAL_BASEDIR}/cpp/context/tensealcontext.cpp
+    ${TENSEAL_BASEDIR}/cpp/context/sealcontext.cpp
+    ${TENSEAL_BASEDIR}/cpp/tensors/bfvvector.cpp
+    ${TENSEAL_BASEDIR}/cpp/tensors/ckksvector.cpp
+    ${TENSEAL_BASEDIR}/cpp/tensors/utils/utils.cpp
```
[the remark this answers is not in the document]
bcebere: Those utils are used only by the tensors. They depend on the context.
If I move them there, the context will rely on a library that relies on the context.

There are other generic utils, like serialization/proto/etc, in the utils folder.

third_party/BUILD
```diff
@@ -0,0 +1,6 @@
+diff --git a/third_party/googletest b/third_party/googletest
+--- a/third_party/googletest
++++ b/third_party/googletest
+@@ -1 +1 @@
+-Subproject commit cb44c86c1aaf31b26797728e93acc950c01dbd5b
++Subproject commit cb44c86c1aaf31b26797728e93acc950c01dbd5b-dirty
```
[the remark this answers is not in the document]
bcebere: fixed

third_party/seal.BUILD
bcebere: Thank for noticing this, I forgot about it. I will fix it.

tenseal/sealapi/sealapi_helpers.cpp
bcebere: yes, I reverted that.

PR #117, opened 2020-07-26
Convolution using im2col technique

philomath213:
## Description
- Computing convolution layers using `im2col` (Image Block to Column) technique.
- Implementing encrypted matrix multiplication, which has some limitation, it requires that the matrix is encoded in vertical scan (encode the matrix transpose instead), then do multiplication with a plain vector. It's a single input single output implementation.
- In order to compute multiple convolution layers, a communication between the server and the client is required in order to apply `im2col` on the input for the next layer.
- This implementation is inspired from PyTorch Convolution implementation, you can read more about it [here](https://towardsdatascience.com/how-are-convolutions-actually-performed-under-the-hood-226523ce7fbf).

## Affected Dependencies
None

## How has this been tested?
- For the moment, I compared the results with torch conv2d only.
EDIT:
- Tests in python.

## Checklist


tests/python/tenseal/tensors/test_ckks_vector.py
```diff
@@ -1,8 +1,12 @@
-import tenseal as ts
-import pytest
-import numpy as np
 import copy
 import pickle
+import math
+import pytest
+
+import numpy as np
+from skimage.util.shape import view_as_windows
```
bcebere, line 7: The Bazel build will fail because of missing depends.

To fix that, you can add the line 
```
requirement("scikit-image")
```

to the tests/python/tenseal/BUILD file

philomath213: I was looking for it.
Got it, thanks.

youben11: So there is no way to have BAZEL fetch that from requirements files?

tenseal/cpp/tensors/ckksvector.cpp
```diff
@@ -570,6 +570,60 @@ CKKSVector& CKKSVector::polyval_inplace(const vector<double>& coefficients) {
...
+    CKKSVector new_vec = *this;
+    new_vec.conv2d_im2col_inplace(kernel, windows_nb);
+    return new_vec;
+}
+
+CKKSVector& CKKSVector::conv2d_im2col_inplace(const vector<double>& kernel,
+                                              size_t windows_nb) {
+    vector<double> plain_vec;
+    size_t chunck_size = windows_nb;
+    size_t chuncks_nb = kernel.size();
```
bcebere, line 584: maybe throw an exception if the kernel si empty or windows_nb == 0?

tenseal/cpp/tensors/ckksvector.cpp
```diff
@@ -570,6 +570,60 @@ CKKSVector& CKKSVector::polyval_inplace(const vector<double>& coefficients) {
...
+                                     size_t windows_nb) {
+    CKKSVector new_vec = *this;
+    new_vec.conv2d_im2col_inplace(kernel, windows_nb);
+    return new_vec;
+}
+
+CKKSVector& CKKSVector::conv2d_im2col_inplace(const vector<double>& kernel,
+                                              size_t windows_nb) {
+    vector<double> plain_vec;
+    size_t chunck_size = windows_nb;
```
bcebere, line 583: small typo: "chunk"

tenseal/cpp/tensors/ckksvector.cpp
```diff
@@ -570,6 +570,60 @@ CKKSVector& CKKSVector::polyval_inplace(const vector<double>& coefficients) {
...
+
+    if (this->_size / windows_nb != chuncks_nb) {
+        throw invalid_argument("Matrix shape doesn't match with vector size");
+    }
+
+    plain_vec.reserve(this->_size);
+
+    for (size_t i = 0; i < chuncks_nb; i++) {
+        vector<double> tmp(chunck_size, kernel[i]);
+        plain_vec.insert(plain_vec.end(), tmp.begin(), tmp.end());
```
bcebere, line 599: ```plain_vec.push_back``` might be more clear here.

tenseal/cpp/tensors/ckksvector.cpp
```diff
@@ -570,6 +570,60 @@ CKKSVector& CKKSVector::polyval_inplace(const vector<double>& coefficients) {
...
+
+    this->mul_plain_inplace(plain_vec);
+
+    auto galois_keys = this->context->galois_keys();
+
+    CKKSVector tmp = *this;
+
+    while (chuncks_nb > 1) {
+        tmp = *this;
+        chuncks_nb = 1 << (static_cast<size_t>(ceil(log2(chuncks_nb))) - 1);
```
bcebere, line 616: For finding the previous power of 2, maybe 
```
uint32_t flp2 (uint32_t x)
{
    x = x | (x >> 1);
    x = x | (x >> 2);
    x = x | (x >> 4);
    x = x | (x >> 8);
    x = x | (x >> 16);
    return x - (x >> 1);
}
```
is faster than the log version.

tests/python/tenseal/tensors/test_ckks_vector.py
bcebere: It actually fetches those requirements, but uses them only in the specified libraries/tests. It won't add any additional dependency by default

tenseal/cpp/tensors/ckksvector.h
```diff
@@ -122,6 +122,13 @@ class CKKSVector {
     CKKSVector polyval(const vector<double>& coefficients);
     CKKSVector& polyval_inplace(const vector<double>& coefficients);
 
+    /*
+    Image Block to Columns .
+    */
+    CKKSVector conv2d_im2col(const vector<double>& kernel, size_t windows_nb);
+    CKKSVector& conv2d_im2col_inplace(const vector<double>& kernel,
+                                      size_t windows_nb);
+
```
youben11, line 131: I think it's more like a convolution with a special input, can you specify convolution and how the input should be encoded?

tenseal/cpp/tensors/ckksvector.cpp
youben11, line 583: I think we can simplify the API by passing a real kernel (matrix) and pad it inside the function, based on the chunk_size.

youben11: Variable names are also confusing, chunk_size and windows_nb seems to be the same thing, why are we using both of them here, what does the API should have instead?

philomath213: Yeah, you are right, we simplify the API in that way, I'm planning to do that in another PR, as well as padding and applying im2col on the input matrix.

tenseal/cpp/tensors/ckksvector.cpp
philomath213: Actually, this will replicate each element in kernel vector for
e.g. if windows_nb == 4 and kernel == [1,2,3,4] => plain_vec == [1,1,1,1, 2,2,2,2, 3,3,3,3, 4,4,4,4]

I found out that this is the faster way to do that, is it correct?

tenseal/cpp/tensors/ckksvector.h
youben11: @philomath213 :point_up:

PR #121, opened 2020-07-28
Drop pickle serialization

bcebere:
## Description
Tensor serialization requires ser/des for the TenSEAL context and for the actual vector.
For performance improvements, we will send the context buffer separately from the tensors, and each tensor deserialization will use a shared pointer to the same context.

Unfortunately, this scenario is not possible with pickle - it forces us to serialize the context every time too - and we will drop the support. We offer a handful of alternatives instead, like our own serialize/deserialize methods or copy/deepcoy support.

## Checklist


youben11:
Can you please cherry pick the last commit into a new branch derived from master, to have a clear file change to review

Issue #118, opened 2020-07-27
Implement Thread Pool


youben11:
## Description
As suggested by @bcebere in #115 , parallel computation using threads shouldn't create them from scratch every time, but instead should assign jobs to a pool of available threads.

We need to add a thread pool system in tenseal/utils as well as integrate it in the vector-matrix multiplication operation.

bcebere:
I can work on this one

PR #124, opened 2020-07-29
Threadpool support for matrix multiplication

bcebere:
## Description
The current flow spawns a number of threads on every multiplication.This can lead to denial of service, we should have a controlled way of creating new threads.

For that, this PR extends the current approach with:

- [x] A blocking queue and a thread pool implementation, for executing the jobs.
- [x] Move the thread creating responsability in the TenSEAL context. 
- [x] Rework the multiplication with globals/mutexes by spliting the input in batches  and using C++'s future promises in the main thread.
- [x] Added Google benchmark for measuring the performance per operations.  
- [x] TODO: Comment the code.

fixes https://github.com/OpenMined/TenSEAL/issues/118

## Checklist


tenseal/cpp/context/tensealcontext.h
```diff
@@ -222,6 +234,11 @@ class TenSEALContext {
      * @returns true if the contexts are identical.
      **/
     bool equals(const std::shared_ptr<TenSEALContext>& other);
+    /**
+     * @returns a pointer to the threadpool dispatcher
+     **/
+    shared_ptr<sync::ThreadPool> dispatcher() { return _dispatcher; }
+    size_t get_concurrency() { return _threads.value_or(1); }
```
youben11, line 241: So the default won't be to check hardware concurrency?

tenseal/cpp/tensors/ckksvector.h
```diff
@@ -109,10 +109,8 @@ class CKKSVector {
     /**
      * Matrix multiplication operations.
      **/
-    CKKSVector matmul_plain(const vector<vector<double>>& matrix,
-                            uint n_threads = 0);
-    CKKSVector& matmul_plain_inplace(const vector<vector<double>>& matrix,
-                                     uint n_threads = 0);
+    CKKSVector matmul_plain(const vector<vector<double>>& matrix);
+    CKKSVector& matmul_plain_inplace(const vector<vector<double>>& matrix);
```
youben11, line 113: Even if now we have a thread pool with a certain number of threads, can't we control the number of jobs created for matmul? like the number of threads is n, and someone can specify a number j <= n of jobs? in the API it will look something more like `vec.matmul(matrix, n_jobs=j)`. Do you think this can be useful, I don't know...

tenseal/cpp/context/tensealcontext.h
bcebere: I'll fix this, I mixed them up

tenseal/cpp/tensors/ckksvector.h
bcebere: Sure, that should be easy to add.

tenseal/cpp/tensors/utils/matrix_ops.h
```diff
@@ -149,21 +102,29 @@ Ciphertext diagonal_ct_vector_matmul_parallel(
...
+    for (size_t i = 0; i < thread_cnt; i++) {
+        future_results.push_back(tenseal_context->dispatcher()->enqueue_task(
+            worker_func, i * batch_size,
+            std::min((i + 1) * batch_size, vector_size)));
+    }
+
+    for (size_t i = 0; i < tenseal_context->get_concurrency(); i++) {
+        tenseal_context->evaluator->add_inplace(result,
+                                                future_results[i].get());
+    }
```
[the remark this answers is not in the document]
bcebere: So, you're saying that this loop might get stuck at 0, let's say, waiting for a slow thread, while the others finished the job, right?

Windows has a mechanism for waiting for multiple objects efficiently, unfortunately I cannot find something similar in standard C++ library, but I'll research it. I was looking for a solution too here, it's not ideal to loop over the futures and do a blocking wait. 

I could experiment with more designs. One other idea was to add the results to a blocking queue, and notify the main thread using a condition variable. That might schedule the operations more efficiently.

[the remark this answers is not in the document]
youben11: The current design is great, I don't think we should jump directly into a different one, I just wanted to give it some thoughts. But this is definitely something I would merge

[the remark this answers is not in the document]
bcebere: I'll make a test with a condition variable&mutex, to see how the benchmarks look that way.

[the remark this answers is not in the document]
bcebere: I did a test using a mutex and a condition variable to notify the main thread(instead of waiting for a std::future), but the benchmarks look pretty similar.

I think it needs better load testing here, I am not sure how to simulate the scenario.

Maybe a random sleep in each worker might show the performance differences(cond variable vs future.get() )
I have the git stash anyway, I can try to add a test in another PR.

tenseal/binding.cpp
```diff
@@ -134,10 +134,14 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
...
-        .def("mm", &CKKSVector::matmul_plain, py::arg("matrix"))
-        .def("mm_", &CKKSVector::matmul_plain_inplace, py::arg("matrix"))
+        .def("matmul", &CKKSVector::matmul_plain, py::arg("matrix"),
+             py::arg("n_batches") = 0)
+        .def("matmul_", &CKKSVector::matmul_plain_inplace, py::arg("matrix"),
+             py::arg("n_batches") = 0)
+        .def("mm", &CKKSVector::matmul_plain, py::arg("matrix"),
+             py::arg("n_batches") = 0)
+        .def("mm_", &CKKSVector::matmul_plain_inplace, py::arg("matrix"),
+             py::arg("n_batches") = 0)
```
youben11: I know this sounds philosophic lol but I think the user doesn't need to know that we process this in batches or not, the batch thing is more specific for our matmul implementation, it can be the parameter name for diagonal_vector_matmul function for example, but the user should only get that the operation is done via `n` jobs, that's why I would prefer the job terminology exposed to the user in the API

PR #134, opened 2020-08-06
Implement im2col and update conv2d

philomath213:
## Description
- Implement matrix vertical and horizontal scan.
- Implement im2col (Image Block to Columns) in C++.
- Update conv2d_im2col implementation in order to simplify the API.
- Implement `im2col_encoding`, it will allow user apply a im2col on input matrix and compute the convolution layer using on TenSEAL, there's no need to pad the input and the kernel nor flatting them in Python.
- Update tests.
- Update Convolution notebook.

## Affected Dependencies
None.

## How has this been tested?
- Unit tests.

## Checklist

### DIFF
--- a/tenseal/binding.cpp
+++ b/tenseal/binding.cpp
@@ -8,6 +8,7 @@
 #include "tenseal/cpp/context/tensealcontext.h"
 #include "tenseal/cpp/tensors/bfvvector.h"
 #include "tenseal/cpp/tensors/ckksvector.h"
+#include "tenseal/cpp/tensors/utils/utils.h"
 
 using namespace tenseal;
 using namespace seal;
@@ -80,6 +81,20 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
         .def("__copy__", [](const BFVVector &self) { return self.deepcopy(); })
         .def("__deepcopy__",
              [](const BFVVector &self, py::dict) { return self.deepcopy(); });
+
+    // CKKSVector utils
+    m.def("im2col_encoding",
+          [](shared_ptr<TenSEALContext> ctx, vector<vector<double>> &input,
+             const size_t kernel_n_rows, const size_t kernel_n_cols,
+             const size_t stride) {
+              vector<vector<double>> view_as_window;
+              vector<double> final_vector;
+              size_t windows_nb = im2col(input, view_as_window, kernel_n_rows,
+                                         kernel_n_cols, stride);
+              vertical_scan(view_as_window, final_vector);
+              CKKSVector ckks_vector = CKKSVector(ctx, final_vector);
+              return make_pair(ckks_vector, windows_nb);
+          });
 
     py::class_<CKKSVector>(m, "CKKSVector")
         // specifying scale
--- a/tenseal/cpp/tensors/ckksvector.cpp
+++ b/tenseal/cpp/tensors/ckksvector.cpp
@@ -564,39 +564,47 @@ CKKSVector& CKKSVector::polyval_inplace(const vector<double>& coefficients) {
     return *this;
 }
 
-CKKSVector CKKSVector::conv2d_im2col(const vector<double>& kernel,
-                                     size_t windows_nb) {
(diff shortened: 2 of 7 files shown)


tenseal/cpp/tensors/utils/utils.h
```diff
@@ -10,6 +12,104 @@ namespace tenseal {
...
+template <typename T>
+void horizontal_scan(const vector<vector<T>>& src, vector<T>& dst) {
+    size_t in_height = src.size();
+    size_t in_width = src[0].size();
+
+    dst.resize(in_height * in_width);
+
+    // check if each row size is equals to in_width
+    if (any_of(src.begin(), src.end(),
+               [in_width](vector<T> i) { return i.size() != in_width; })) {
```
bcebere, line 27: const ref

tenseal/cpp/tensors/utils/utils.h
```diff
@@ -10,6 +12,104 @@ namespace tenseal {
...
+template <typename T>
+void vertical_scan(const vector<vector<T>>& src, vector<T>& dst) {
+    size_t in_height = src.size();
+    size_t in_width = src[0].size();
+
+    dst.resize(in_height * in_width);
+
+    // check if each row size is equals to in_width
+    if (any_of(src.begin(), src.end(),
+               [in_width](vector<T> i) { return i.size() != in_width; })) {
```
bcebere, line 52: const ref

tenseal/cpp/tensors/utils/utils.h
```diff
@@ -10,6 +12,104 @@ namespace tenseal {
 using namespace seal;
 using namespace std;
 
+/**
+ * horizontally scan matrix (vector of vectors)
+ **/
+template <typename T>
+void horizontal_scan(const vector<vector<T>>& src, vector<T>& dst) {
```
bcebere, line 19: These operations might be easier to execute using a gsl::multi_span structure. For sure not for this PR, I don't know if they actually help. But they should prevent extra-copies

philomath213: Yeah, I will check this out, maybe they will be helpful.

tenseal/binding.cpp
```diff
@@ -81,6 +82,20 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
...
+             const size_t stride) {
+              vector<vector<double>> view_as_window;
+              vector<double> final_vector;
+              size_t windows_nb = im2col(input, view_as_window, kernel_n_rows,
+                                         kernel_n_cols, stride);
+              vertical_scan(view_as_window, final_vector);
+              CKKSVector ckks_vector = CKKSVector(ctx, final_vector);
+              return make_pair(ckks_vector, windows_nb);
+          });
+
```
youben11, line 98: Maybe not in this PR, but the encoding shouldn't be doing encryption, just encoding it into a vector, then encryption should be something separate

tenseal/cpp/tensors/ckksvector.cpp
```diff
@@ -564,39 +564,47 @@ CKKSVector& CKKSVector::polyval_inplace(const vector<double>& coefficients) {
...
     if (windows_nb == 0) {
         throw invalid_argument("Windows number can't be zero");
     }
 
-    if (kernel.empty()) {
-        throw invalid_argument("Kernel vector can't be empty");
+    if (kernel.empty() ||
+        (any_of(kernel.begin(), kernel.end(),
+                [](const vector<double>& i) { return i.empty(); }))) {
+        throw invalid_argument("Kernel matrix can't be empty");
```
youben11, line 583: What about something like this? to check whether the vector<vector> is a well organized matrix?

```suggestion
    if (kernel.empty() ||
        (any_of(kernel.begin(), kernel.end(),
                [](const vector<double>& i) { return i.size() == kernel[0].size(); }))) {
        throw invalid_argument("Kernel matrix can't be empty");
```

philomath213: The test will be wrong in the case where `kernel[0].size() == 0`.
in other hand `horizontal_scan` always check if the rows have the same size.

PR #139, opened 2020-08-17
Checkout with submodules

philomath213:
## Description
- Use actions/checkout@v2 instead of v1.
- Set submodules argument in actions/checkout, this will make sure to initialize and update recursively any submodules.
- Remove submodules initialize and updating in scripts and docker files.
- Format workflows YML files.

## Affected Dependencies
None.

## How has this been tested?
CI workflow will make sure that everything is working fine.

## Checklist

### DIFF
--- a/.github/workflows/scripts/build_nix.sh
+++ b/.github/workflows/scripts/build_nix.sh
@@ -1,6 +1,4 @@
 #!/bin/sh
 
-# fetch third_party libraries
-git submodule init && git submodule update
 # we need to install tenseal in order to build the cpp library
 pip install .
--- a/.github/workflows/scripts/build_windows.bat
+++ b/.github/workflows/scripts/build_windows.bat
@@ -6,8 +6,6 @@ set PATH=%PATH%;%programfiles%/protobuf/bin/;%programfiles(x86)%/protobuf/bin/
 :: Protubf library enforces the MT flag on the static .lib build. If we want to link it, we need to use MT everywhere, otherwise the build will fail.
 :: The general recommendation is to use MD, but for that we need to build and use protobufs as DLL.
 set _CL_=/MT
-
-git submodule init && git submodule update
  
 msbuild .\third_party\SEAL\native\src\SEAL.vcxproj /p:configuration=Release /p:platform=x64 /p:SolutionDir=..\..\
 
--- a/.github/workflows/scripts/release_macos.sh
+++ b/.github/workflows/scripts/release_macos.sh
@@ -1,6 +1,4 @@
 #!/bin/sh
 
-# fetch third_party libraries
-git submodule init && git submodule update
 pip wheel . -w wheel/
 twine upload wheel/*
--- a/.github/workflows/scripts/release_ubuntu.sh
+++ b/.github/workflows/scripts/release_ubuntu.sh
@@ -1,8 +1,5 @@
 #!/bin/sh
 
-# fetch third_party libraries
-git submodule init && git submodule update
-
 pip wheel . -w dist/ --no-deps
 auditwheel repair dist/*.whl --plat $AUDITWHEEL_PLAT
 twine upload --skip-existing wheelhouse/*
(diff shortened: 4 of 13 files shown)


.github/workflows/scripts/build_nix.sh
```diff
@@ -1,6 +1,4 @@
 #!/bin/sh
 
-# fetch third_party libraries
-git submodule init && git submodule update
```
youben11, line 4: I think we can leave this here, as we may want to refer to this script for users to try

.github/workflows/scripts/build_windows.bat
```diff
@@ -6,8 +6,6 @@ set PATH=%PATH%;%programfiles%/protobuf/bin/;%programfiles(x86)%/protobuf/bin/
 :: Protubf library enforces the MT flag on the static .lib build. If we want to link it, we need to use MT everywhere, otherwise the build will fail.
 :: The general recommendation is to use MD, but for that we need to build and use protobufs as DLL.
 set _CL_=/MT
-
-git submodule init && git submodule update
```
youben11, line 10: same

.github/workflows/scripts/release_macos.sh
```diff
@@ -1,6 +1,4 @@
 #!/bin/sh
 
-# fetch third_party libraries
-git submodule init && git submodule update
```
youben11, line 4: same

.github/workflows/scripts/release_ubuntu.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/sh
 
-# fetch third_party libraries
-git submodule init && git submodule update
-
```
youben11, line 5: same

docker-images/Dockerfile-py36
```diff
@@ -6,8 +6,6 @@ RUN apt-get update
 RUN apt-get install -y git build-essential cmake libprotobuf-dev protobuf-compiler
 COPY ./ ./TenSEAL
 WORKDIR /TenSEAL
-# 3rd party libraries
-RUN git submodule init && git submodule update
```
youben11, line 10: Why removing from here? this isn't affected by the github action checkout

docker-images/Dockerfile-py37
```diff
@@ -6,8 +6,6 @@ RUN apt-get update
 RUN apt-get install -y git build-essential cmake libprotobuf-dev protobuf-compiler
 COPY ./ ./TenSEAL
 WORKDIR /TenSEAL
-# 3rd party libraries
-RUN git submodule init && git submodule update
```
youben11, line 10: same

docker-images/Dockerfile-py38
```diff
@@ -6,8 +6,6 @@ RUN apt-get update
 RUN apt-get install -y git build-essential cmake libprotobuf-dev protobuf-compiler
 COPY ./ ./TenSEAL
 WORKDIR /TenSEAL
-# 3rd party libraries
-RUN git submodule init && git submodule update
```
youben11, line 10: same

.github/workflows/scripts/build_nix.sh
philomath213: Build instructions are specified and explained in README, I think users shouldn't refer to workflows scripts (both build and release) since they are related to GitHub Actions environment.

docker-images/Dockerfile-py36
youben11: ping @philomath213

philomath213: Same for other scripts, docker-image workflow will checkout the submodules, and Dockerfile is using `COPY ./ ./TenSEAL` to copy the content to the build container. so there's is no need to to do that here.
One more thing we can't build a docker image with different submodules version, if we checkout a submodule manually to a different header, the command `git submodule update` will checkout back to specified header in our repository.

PR #156, opened 2020-11-11
CKKS Tensor

youben11:
## Description
This introduce an encrypted tensor using CKKS. It can stores an N-dimensional plain tensor into a (N-1)-dimensional encrypted tensor by batching one of the axis into the ciphertexts.

## Checklist

### DIFF
--- a/tenseal/cpp/tensors/ckkstensor.cpp
+++ b/tenseal/cpp/tensors/ckkstensor.cpp
@@ -1,0 +1,47 @@
+#include "tenseal/cpp/tensors/ckkstensor.h"
+
+using namespace seal;
+using namespace std;
+
+namespace tenseal {
+
+CKKSTensor::CKKSTensor(const shared_ptr<TenSEALContext>& ctx,
+                       const vector<double>& data, std::optional<double> scale)
+    : _shape({data.size()}), _strides({1}) {
+    this->link_tenseal_context(ctx);
+    if (scale.has_value()) {
+        this->_init_scale = scale.value();
+    } else {
+        this->_init_scale = ctx->global_scale();
+    }
+
+    for (int i = 0; i < data.size(); i++)
+        _data.push_back(CKKSTensor::encrypt(ctx, this->_init_scale, data[i]));
+}
+
+Ciphertext CKKSTensor::encrypt(const shared_ptr<TenSEALContext>& ctx,
+                               const double scale, const double data) {
+    return encrypt(ctx, scale, {data});
+}
+
+Ciphertext CKKSTensor::encrypt(const shared_ptr<TenSEALContext>& ctx,
+                               const double scale, const vector<double>& data) {
+    if (data.empty()) {
+        throw invalid_argument("Attempting to encrypt an empty vector");
+    }
+    auto slot_count = ctx->slot_count<CKKSEncoder>();
+    if (data.size() > slot_count)
+        // number of slots available is poly_modulus_degree / 2
+        throw invalid_argument(
+            "can't encrypt vectors of this size, please use a larger "
+            "polynomial modulus degree.");
+
+    Ciphertext ciphertext(ctx->seal_context());
+    Plaintext plaintext;
+    ctx->encode<CKKSEncoder>(data, plaintext, scale);
+    ctx->encryptor->encrypt(plaintext, ciphertext);
+
+    return ciphertext;
+}
+
+}  // namespace tenseal
--- a/tenseal/cpp/tensors/ckkstensor.cpp
+++ b/tenseal/cpp/tensors/ckkstensor.cpp
@@ -176,10 +176,12 @@ string CKKSTensor::save() const {
 
 shared_ptr<CKKSTensor> CKKSTensor::copy() const {
     // TODO
+    return nullptr;
 }
 
 shared_ptr<CKKSTensor> CKKSTensor::deepcopy() const {
     // TODO
+    return nullptr;
 }
 
 }  // namespace tenseal
--- a/tenseal/cpp/tensors/ckkstensor.cpp
+++ b/tenseal/cpp/tensors/ckkstensor.cpp
@@ -18,13 +18,8 @@ CKKSTensor::CKKSTensor(const shared_ptr<TenSEALContext>& ctx,
 
     vector<double> plain_data = tensor.data();
     for (int i = 0; i < plain_data.size(); i++)
-        _data.push_back(
-            CKKSTensor::encrypt(ctx, this->_init_scale, plain_data[i]));
-}
-
-Ciphertext CKKSTensor::encrypt(const shared_ptr<TenSEALContext>& ctx,
-                               const double scale, const double data) {
-    return encrypt(ctx, scale, vector<double>({data}));
+        _data.push_back(CKKSTensor::encrypt(ctx, this->_init_scale,
+                                            vector<double>({plain_data[i]})));
 }
 
 Ciphertext CKKSTensor::encrypt(const shared_ptr<TenSEALContext>& ctx,
--- a/tenseal/cpp/tensors/ckkstensor.cpp
+++ b/tenseal/cpp/tensors/ckkstensor.cpp
@@ -1,13 +1,14 @@
 #include "tenseal/cpp/tensors/ckkstensor.h"
 
+namespace tenseal {
+
 using namespace seal;
 using namespace std;
 
-namespace tenseal {
-
 CKKSTensor::CKKSTensor(const shared_ptr<TenSEALContext>& ctx,
-                       const vector<double>& data, std::optional<double> scale)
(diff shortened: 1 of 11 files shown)


tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -0,0 +1,47 @@
+#include "tenseal/cpp/tensors/ckkstensor.h"
+
+using namespace seal;
+using namespace std;
+
+namespace tenseal {
+
+CKKSTensor::CKKSTensor(const shared_ptr<TenSEALContext>& ctx,
+                       const vector<double>& data, std::optional<double> scale)
```
bcebere, line 9: Why isn't data a PlainTensor directly?

tenseal/cpp/tensors/ckkstensor.h
```diff
@@ -0,0 +1,87 @@
...
+   public:
+    
+    /**
+     * Create a new CKKSTensor from an 1D vector.
+     * @param[in] input vector.
+     * @param[in] input vector.
+     * @param[in] input vector.
+     */
+    static shared_ptr<CKKSTensor> Create(const shared_ptr<TenSEALContext>& ctx,
+                                         const vector<double>& data,
```
bcebere, line 21: Maybe PlainTensor instead of vector?

tenseal/cpp/tensors/ckkstensor.cpp
youben11: True that expecting a PlainTensor would make life much easier. I will only encrypt data and use the previously saved strides and shape.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -0,0 +1,185 @@
...
+    if (scale.has_value()) {
+        this->_init_scale = scale.value();
+    } else {
+        this->_init_scale = ctx->global_scale();
+    }
+
+    vector<double> plain_data = tensor.data();
+    for (int i = 0; i < plain_data.size(); i++)
+        _data.push_back(
+            CKKSTensor::encrypt(ctx, this->_init_scale, plain_data[i]));
```
bcebere, line 22: It's for another PR, but I think it could be useful to add iterator support for the PlainTensors.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+Ciphertext CKKSTensor::encrypt(const shared_ptr<TenSEALContext>& ctx,
+                               const double scale, const double data) {
+    return encrypt(ctx, scale, vector<double>({data}));
+}
+
+Ciphertext CKKSTensor::encrypt(const shared_ptr<TenSEALContext>& ctx,
+                               const double scale, const vector<double>& data) {
```
bcebere, line 31: data can be a PlainTensor as well.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    Ciphertext ciphertext(ctx->seal_context());
+    Plaintext plaintext;
+    ctx->encode<CKKSEncoder>(data, plaintext, scale);
+    ctx->encryptor->encrypt(plaintext, ciphertext);
+
+    return ciphertext;
+}
+
+PlainTensor<double> CKKSTensor::decrypt() const {
```
bcebere, line 50: This call can be moved to the encrypted_tensor interface, it's just duplicated code, I guess.

(for another PR)

bcebere (commented):
Great work, this is shaping up nicely.

For the save/load part you will need to update the protobuffers as well.
If you think the current encrypted tensor API is hard to extend, we can rethink the interface.

tenseal/tensors/__init__.py
```diff
@@ -103,4 +120,32 @@ def ckks_vector_from(context, data):
     )
 
 
-__all__ = ["bfv_vector", "bfv_vector_from", "ckks_vector", "ckks_vector_from"]
+def ckks_tensor(context, tensor, scale=None):
+    """Constructor method for the CKKSVector object, which can store a list
```
bcebere, line 124: CKKSTensor instead of CKKSVector?

tenseal/tensors/__init__.py
```diff
@@ -103,4 +120,32 @@ def ckks_vector_from(context, data):
...
+    of float numbers in encrypted form, using the CKKS homomorphic encryption
+    scheme.
+
+    Args:
+        context: a TenSEALContext object, holding the encryption parameters and keys.
+        tensor: a PlainTensorDouble holding data to be encrypted.
+        scale: the scale to be used to encode vector values. CKKSVector will use the global_scale provided by the context if it's set to None.
```
bcebere, line 131: CKKSTensor?

tests/python/tenseal/test_enc_dec.py
```diff
@@ -94,3 +96,16 @@ def test_ckks_empty_encryption(plain_vec):
     with pytest.raises(ValueError) as e:
         ckks_vec = ts.ckks_vector(context, plain_vec, scale)
     assert str(e.value) == "Attempting to encrypt an empty vector"
+
+
```
bcebere, line 100: Could be possible to reuse the CKKSVector tests?

bcebere (commented):
Looking good so far, just a few minor observations

tests/python/tenseal/test_enc_dec.py
youben11: I just wanted to test this simple functionality for now, and the CKKSVector tests aren't using plainvector objects and only takes vectors as input, which should be N-D for the CKKSTensor tests. So maybe they will need to be separate at some point.

tenseal/cpp/tensors/ckkstensor.cpp
bcebere: The rest of the API uses PlainTensor for encrypt. Maybe we should do the same here.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -0,0 +1,187 @@
...
+        this->_init_scale = ctx->global_scale();
+    }
+
+    vector<double> plain_data = tensor.data();
+    for (int i = 0; i < plain_data.size(); i++)
+        _data.push_back(
+            CKKSTensor::encrypt(ctx, this->_init_scale, plain_data[i]));
+}
+
+Ciphertext CKKSTensor::encrypt(const shared_ptr<TenSEALContext>& ctx,
```
bcebere, line 25: Do we need a dedicated call for vectors of length 1?

The other implementations don't have it, as far as I remember.

bcebere (changes requested):
There are still a lot of TODOs in the code, a hope we do not forget about them.

tenseal/cpp/tensors/ckkstensor.cpp
youben11: Not of length 1, but we will need to encrypt vectors when we use batching, and I wanted to use the same function for encrypting a single value, which can be just a vector of length one. Is there a better way to do this?

tenseal/cpp/tensors/ckkstensor.cpp
youben11: Yeah, that would be great!

tenseal/cpp/tensors/ckkstensor.cpp
youben11: I see this encrypt more as an internal private function, that takes a vector of double values to batch into a single ciphertext. The constructor is responsible for manipulating the PlainTensor and constructing these vectors for encryption. What do you think?

tenseal/cpp/tensors/ckkstensor.cpp
bcebere: I was just wondering why we need a different call.

We already have encrypt(const shared_ptr<TenSEALContext>& ctx, const double scale, const vector<double>& data)

tenseal/cpp/tensors/ckkstensor.cpp
bcebere: I think it would be better for consistency to work with PlainTensors, because the rest of the tensors work with it already.

Or we should update them as well, to have a similar encrypt declaration.

tenseal/cpp/tensors/ckkstensor.cpp
youben11: Got you! Yeah you are right!

bcebere (approved):
LGTM! 

One idea: Leave some issues on Github with improvements/tests left to do for CKKSTensor; maybe others can help with them.

youben11:
@bcebere Yeah, all the TODOs will be translated into separate issues, that was the goal of doing an early merge.

PR #158, opened 2020-11-22
SEAL 3.6

bcebere:
## Description
SEAL 3.6

Changelog: https://github.com/microsoft/SEAL/releases/tag/v3.6.0

## Checklist


CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+include_directories(third_party/SEAL/native/src)
 
 # ##############################################################################
 # Third party
```
youben11, line 12: We can remove the comments here as we no longer build it using VS, right?

tenseal/cpp/context/sealcontext.cpp
```diff
@@ -28,16 +28,15 @@ EncryptionParameters create_bfv_parameters(size_t poly_modulus_degree,
...
         CoeffModulus::Create(poly_modulus_degree, coeff_mod_bit_sizes));
 
     return parameters;
 }
 
-shared_ptr<seal::SEALContext> create_context(EncryptionParameters parms) {
-    auto context = SEALContext::Create(parms);
-    return context;
+SEALContext create_context(EncryptionParameters parms) {
+    return SEALContext(parms);
```
youben11, line 40: Is there any reason for this change?

tenseal/cpp/context/tensealencoder.h
```diff
@@ -115,7 +115,7 @@ class TenSEALEncoder {
     /*
     Stores a shared_pointer to the SEAL Context.
     */
-    std::shared_ptr<SEALContext> _context;
+    SEALContext _context;
```
youben11, line 118: So we now keep a full copy of the context instead of a pointer? Why is that?

youben11 (changes requested):
Looks great! I left some questions here that we can discuss later over the meeting

tenseal/cpp/context/sealcontext.cpp
bcebere: SEALContext::Create was removed in this release

tenseal/cpp/context/tensealencoder.h
bcebere: They dropped the Create call, which created the shared_ptr, and they are copying the SealContexts everytime.

Actually, I kept the shared_ptr in the TenSEAL context, we can do that here as well.

tenseal/cpp/context/sealcontext.cpp
bcebere: This is their commit
https://github.com/microsoft/SEAL/commit/a94374d348fb28390e6191cbed52830d67617736#diff-797d0099b848a1dbe9380766c6034f9f500c85b708ec931ea6dc7aa9520f3767

tenseal/cpp/context/tensealencoder.h
youben11: Yeah, I guess it doesn't require a full copy, as we will be mainly accessing attributes, or do you have a reason to believe we shouldn't use a pointer here?

tenseal/cpp/context/sealcontext.cpp
youben11: Thanks! I couldn't find anything about it in the release note, as I was wondering if it's only a design choice, or that there is implications in having multiple holders of the same context.

tenseal/cpp/context/tensealencoder.h
bcebere: I will revert it here. Initially, I was going to drop std::shared_ptr<SEALContext> everywhere, and this is some partial change.

tenseal/cpp/context/sealcontext.cpp
bcebere: While is not great to change the API that much, they do not need a shared ptr, their context is pretty light.

youben11 (approved):

Issue #161, opened 2020-11-23
CKKSTensor - add, sub, mul


youben11:
## Feature Description
Add element-wise operations (add, sub, mul) to CKKSTensor. These methods should first check the shapes of the operands to check if they are compatible (same shape or one can be broadcasted), then perform the operation depending on their shapes. These methods should have approximately the same code, at the difference of performing different operations (with mul requiring relin/rescale).

Performing these operation in parallel using the threadpool should be prefered.

## Additional Context
The methods' API is already set.

youben11:
I will work on this

PR #169, opened 2020-11-27
CKKSTensor - add,sub,mul

youben11:
## Description
Implementing the operations described in #161.

## Checklist

### DIFF
--- a/tenseal/cpp/tensors/ckkstensor.cpp
+++ b/tenseal/cpp/tensors/ckkstensor.cpp
@@ -87,7 +87,8 @@ shared_ptr<CKKSTensor> CKKSTensor::add_inplace(
     auto worker_func = [&](vector<Ciphertext>* dst, vector<Ciphertext>* src,
                            size_t start, size_t end) -> bool {
         for (size_t i = start; i < end; i++) {
-            this->tenseal_context()->evaluator->add_inplace((*dst)[i], (*src)[i]);
+            this->tenseal_context()->evaluator->add_inplace((*dst)[i],
+                                                            (*src)[i]);
         }
         return true;
     };
--- a/tenseal/cpp/tensors/ckkstensor.cpp
+++ b/tenseal/cpp/tensors/ckkstensor.cpp
@@ -199,7 +199,16 @@ shared_ptr<CKKSTensor> CKKSTensor::op_inplace(
                     std::min((i + 1) * batch_size, this->_data.size())));
         }
         // waiting
-        for (size_t i = 0; i < futures.size(); i++) futures[i].get();
+        std::optional<std::exception> fail;
+        for (size_t i = 0; i < futures.size(); i++) {
+            try {
+                futures[i].get();
+            } catch (std::exception& e) {
+                fail = e;
+            }
+        }
+
+        if (fail) throw invalid_argument(fail.value().what());
     }
 
     return shared_from_this();
--- a/tenseal/cpp/tensors/ckkstensor.cpp
+++ b/tenseal/cpp/tensors/ckkstensor.cpp
@@ -129,6 +129,25 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
     return shared_from_this();
 }
 
+template <typename Other>
+void CKKSTensor::perform_op(seal::Ciphertext& ct, Other other, OP op) {
+    this->auto_same_mod(other, ct);
+    switch (op) {
+        case OP::ADD:
+            this->tenseal_context()->evaluator->add_inplace(ct, other);
+            break;
+        case OP::SUB:
+            this->tenseal_context()->evaluator->sub_inplace(ct, other);
+        case OP::MUL:
+            this->tenseal_context()->evaluator->multiply_inplace(ct, other);
+            this->auto_relin(ct);
+            this->auto_rescale(ct);
+            break;
+        default:
+            throw invalid_argument("operation not defined");
+    }
+}
+
 shared_ptr<CKKSTensor> CKKSTensor::op_inplace(
     const shared_ptr<CKKSTensor>& operand, OP op) {
     // TODO implement broadcasting
@@ -141,24 +160,7 @@ shared_ptr<CKKSTensor> CKKSTensor::op_inplace(
 
     auto worker_func = [&](size_t start, size_t end) -> bool {
         for (size_t i = start; i < end; i++) {
-            // TODO auto_mod_switch
-            switch (op) {
-                case OP::ADD:
-                    this->tenseal_context()->evaluator->add_inplace(
-                        this->_data[i], operand->_data[i]);
-                    break;
-                case OP::SUB:
-                    this->tenseal_context()->evaluator->sub_inplace(
-                        this->_data[i], operand->_data[i]);
-                case OP::MUL:
-                    this->tenseal_context()->evaluator->multiply_inplace(
-                        this->_data[i], operand->_data[i]);
-                    // TODO: auto_relin auto_rescale
-                    break;
-
-                default:
-                    throw invalid_argument("operation not defined");
-            }
+            this->perform_op(this->_data[i], operand->_data[i], op);
         }
         return true;
     };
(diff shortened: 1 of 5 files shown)


tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -76,7 +76,37 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
 
 shared_ptr<CKKSTensor> CKKSTensor::add_inplace(
     const shared_ptr<CKKSTensor>& to_add) {
-    // TODO
+    // TODO implement broadcasting
+    if (this->_shape != to_add->_shape) {
+        // TODO provide a better message (what are the shapes)
+        throw invalid_argument("Operands doesn't have the same shape");
```
bcebere, line 82: small typo: "don't"

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -76,7 +76,37 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
...
+    }
+
+    size_t n_jobs = this->tenseal_context()->dispatcher_size();
+
+    auto worker_func = [&](vector<Ciphertext>* dst, vector<Ciphertext>* src,
+                           size_t start, size_t end) -> bool {
+        for (size_t i = start; i < end; i++) {
+            this->tenseal_context()->evaluator->add_inplace((*dst)[i], (*src)[i]);
```
bcebere: I don't think this is thread-safe, accessing dst on the workers.

You should add a mutex for dst here.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -76,7 +76,38 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
...
+        // TODO provide a better message (what are the shapes)
+        throw invalid_argument("Operands doesn't have the same shape");
+    }
+
+    size_t n_jobs = this->tenseal_context()->dispatcher_size();
+
+    auto worker_func = [&](vector<Ciphertext>* dst, vector<Ciphertext>* src,
+                           size_t start, size_t end) -> bool {
+        for (size_t i = start; i < end; i++) {
+            this->tenseal_context()->evaluator->add_inplace((*dst)[i],
```
bcebere, line 90: I don;t think this is thread safe, to access dst this way.

You either use a mutex over dst/this->data, or return the batch result, and let the main thread do the computations over this->data.

bcebere (changes requested):

tenseal/cpp/tensors/ckkstensor.cpp
youben11: I thought writing to separate locations should be fine, but it seems the object itself isn't thread safe, thanks for spotting that.

bcebere: Aah, my bad.

Yes, that is thread safe, as long the dst object is allocated, and you won't write in the same slot from multiple threads.

But I'm a bit confused:
 - you are creating the lambda with & ref, so you can access this->data directly.
 - same for to_add, you are creating a ref to the parameters as well.

I was confused by the pointer parameters, I though you reallocate something

youben11: Oh yeah! Initially I used pointers then I started using the lambda with & ref to access the evaluator, but now I can remove the dst and src parameters as I can access them directly.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -76,7 +76,37 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
 
 shared_ptr<CKKSTensor> CKKSTensor::add_inplace(
```
bcebere, line 77: The PR should inplace the add_plain implementation too?

youben11: Yeah, as well as sub and mul. I just wanted to have a first review since other methods would look similar.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -1,5 +1,9 @@
 #include "tenseal/cpp/tensors/ckkstensor.h"
 
+#define OP_ADD 0
```
bcebere, line 3: It is preferred to use enum classes instead of defines or plain enums, to prevent bugs/automatic casts or conversions.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -124,22 +128,177 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
...
+    // TODO batched ops
+    if (this->_shape != operand.shape()) {
+        // TODO provide a better message (what are the shapes)
+        throw invalid_argument("Operands don't have the same shape");
+    }
+
+    size_t n_jobs = this->tenseal_context()->dispatcher_size();
+    auto operand_data = operand.data();
+
+    auto worker_func = [&](size_t start, size_t end) -> bool {
```
bcebere, line 195: Maybe worker_func can be shared between op_plain_inplace overloads.

bcebere (commented):
This already looks great!

There is a bit of code duplication for worker_func, maybe it can be reused somehow.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -129,22 +129,165 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
...
 
+void CKKSTensor::perform_op(seal::Ciphertext& ct, seal::Ciphertext other,
+                            OP op) {
+    this->auto_same_mod(other, ct);
+    switch (op) {
+        case OP::ADD:
+            this->tenseal_context()->evaluator->add_inplace(ct, other);
+            break;
+        case OP::SUB:
+            this->tenseal_context()->evaluator->sub_inplace(ct, other);
```
bcebere, line 140: break; ?

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -129,22 +129,165 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
...
+
+void CKKSTensor::perform_plain_op(seal::Ciphertext& ct, seal::Plaintext other,
+                                  OP op) {
+    this->auto_same_mod(other, ct);
+    switch (op) {
+        case OP::ADD:
+            this->tenseal_context()->evaluator->add_plain_inplace(ct, other);
+            break;
+        case OP::SUB:
+            this->tenseal_context()->evaluator->sub_plain_inplace(ct, other);
```
bcebere, line 159: break;

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -129,22 +129,165 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
...
+    if (this->_shape != operand->_shape) {
+        // TODO provide a better message (what are the shapes)
+        throw invalid_argument("Operands don't have the same shape");
+    }
+
+    size_t n_jobs = this->tenseal_context()->dispatcher_size();
+
+    auto worker_func = [&](size_t start, size_t end) -> bool {
+        for (size_t i = start; i < end; i++) {
+            this->perform_op(this->_data[i], operand->_data[i], op);
```
bcebere, line 183: As far as I see, perform_op can throw exceptions.

We shouldn't not throw exceptions in the workers, as it might lead to undefined behavior in the other threads. I think we had a similar crash here.  https://github.com/OpenMined/TenSEAL/pull/151

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -129,22 +129,165 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
...
+        throw invalid_argument("Operands don't have the same shape");
+    }
+
+    size_t n_jobs = this->tenseal_context()->dispatcher_size();
+    auto operand_data = operand.data();
+
+    auto worker_func = [&](size_t start, size_t end) -> bool {
+        Plaintext plaintext;
+        for (size_t i = start; i < end; i++) {
+            this->tenseal_context()->encode<CKKSEncoder>(
```
bcebere, line 221: The same issue about throwing exceptions in the workers

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -129,22 +129,165 @@ shared_ptr<CKKSTensor> CKKSTensor::power_inplace(unsigned int power) {
...
+shared_ptr<CKKSTensor> CKKSTensor::op_plain_inplace(const double& operand,
+                                                    OP op) {
+    size_t n_jobs = this->tenseal_context()->dispatcher_size();
+    Plaintext plaintext;
+    this->tenseal_context()->encode<CKKSEncoder>(operand, plaintext,
+                                                 this->_init_scale);
+
+    auto worker_func = [&](size_t start, size_t end) -> bool {
+        for (size_t i = start; i < end; i++) {
+            this->perform_plain_op(this->_data[i], plaintext, op);
```
bcebere, line 255: same

bcebere (changes requested):
Looks great, just a few observations!

Regarding workers, we had a crash in the past, in a similar scenario, here is a regression test for ref https://github.com/OpenMined/TenSEAL/blob/master/tests/python/tenseal/tensors/test_regression.py#L6

bcebere (approved):
LGTM! awesome work

Issue #165, opened 2020-11-23
CKKSTensor - sum


youben11:
## Feature Description
These feature should take which axis we should sum over. If it's the batch axis, then we should use ciphertext rotations and additions. Otherwise, usual tensor summation should work.

Performing this operation in parallel using the threadpool should be preferred.

bcebere:
I can work on this one.

Issue #162, opened 2020-11-23
CKKSTensor negate


youben11:
## Feature Description
Implement the negate operation for CKKSTensor. This feature should mainly loop over internal ciphertexts and perform the negate operation

Performing this operation in parallel using the threadpool should be prefered.

PR #171, opened 2020-11-30
CKKSTensor ops

bcebere:
## Description
 - [x] Support sum by any axis.
 - [x] In case of batching, supports sum only by the batching axis. 
 - [x] Fix batching for other dimensions than 0.
 - [x] Add negate operation.
 - [x] Add square operation.
 - [x] gtest+ python tests

fixes https://github.com/OpenMined/TenSEAL/issues/165
fixes https://github.com/OpenMined/TenSEAL/issues/162
fixes https://github.com/OpenMined/TenSEAL/issues/172

## Checklist


tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -188,7 +193,50 @@ shared_ptr<CKKSTensor> CKKSTensor::mul_plain_inplace(const double& to_mul) {
...
+
+            size_t new_idx = 0;
+            for (size_t pidx = 0; pidx < pos.size(); ++pidx)
+                new_idx += new_strides[pidx] * pos[pidx];
+
+            tenseal_context()->evaluator->add_inplace(new_data[new_idx],
+                                                      _data[idx]);
+        }
+        _data = new_data;
+    }
```
youben11, line 237: I think there is an add_many function in SEAL, that takes a vector of ciphertexts. Can it be helpful here? We wouldn't then need to initialize ciphertexts to 0.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -188,7 +193,50 @@ shared_ptr<CKKSTensor> CKKSTensor::mul_plain_inplace(const double& to_mul) {
...
 
 shared_ptr<CKKSTensor> CKKSTensor::sum_inplace(size_t axis) {
-    // TODO
+    if (axis >= _shape.size()) throw invalid_argument("invalid axis");
+
+    auto new_shape = _shape;
+    new_shape.erase(new_shape.begin() + axis);
+
+    if (_batch_size) {
+        if (axis != 0) throw invalid_argument("unsupported argument");
```
youben11, line 202: Why isn't it possible to sum other axis when there is a batch_axis? It should be still possible to do so without affecting the batch_size, keeping it the same while summing up over the other axis

tenseal/tensors/__init__.py
```diff
@@ -120,7 +122,7 @@ def ckks_vector_from(context, data):
     )
 
 
-def ckks_tensor(context, tensor, scale=None):
+def ckks_tensor(context, tensor, scale=None, batch=True):
```
youben11, line 125: I think beginner users might not expect the batching to happen, they might find it weird somehow. Can we make the default to false here?

tenseal/binding.cpp
```diff
@@ -27,6 +28,8 @@ void bind_plain_tensor(py::module &m, const std::string &name) {
         .def("shape", &type::shape)
         .def("strides", &type::strides)
         .def("size", &type::size)
+        .def("batch", &type::batch)
+        .def("__len__", &type::size)
```
youben11, line 32: Most tensor libraries seems to return the shape[0] in len(tensor). Should we implement this the same way?

tenseal/binding.cpp
```diff
@@ -27,6 +28,8 @@ void bind_plain_tensor(py::module &m, const std::string &name) {
         .def("shape", &type::shape)
         .def("strides", &type::strides)
         .def("size", &type::size)
+        .def("batch", &type::batch)
```
youben11, line 31: Can we please have some tests for this feature?

youben11 (changes requested):
This looks amazing!

tenseal/binding.cpp
bcebere: Isn't this already implemented? or I'm missing something

https://github.com/OpenMined/TenSEAL/blob/master/tenseal/cpp/tensors/plain_tensor.h#L233

tenseal/binding.cpp
bcebere: There are already c++ tests https://github.com/OpenMined/TenSEAL/blob/ckks_tensor_improvements/tests/cpp/tensors/plaintensor_test.cpp#L124

I will add python tests as well.

tenseal/binding.cpp
youben11: Oh yeah, my bad.

youben11 (approved):
Amazing! Thanks Bogdan

PR #185, opened 2020-12-06
Python object wrappers

youben11:
## Description
C objects from the TenSEAL library, even binded to python, are hard to deal with in a pythonic way (e.g. syft requires adding attributes dynamically). These PR introduce python classes that wrap low level classes.

## Affected Dependencies
PlainTensor now have a `raw` attribute to access the internal vector storage, which was previously accessed via `data`, which is now used in all wrappers to store the low level objects, in their Python object wrappers.

## How has this been tested?
- Some minor API changes were reflected into the tests, however, the biggest part of the tests ran successfully with this new changes, which means that the new Python wrappers behave similarly to the C objects.

## Checklist
- [ ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)


tenseal/tensors/abstract_tensor.py
```diff
@@ -0,0 +1,152 @@
...
+            return other.data
+        else:
+            try:
+                other = ts.plain_tensor(other, dtype=dtype)
+                other = other.data
+            except TypeError:
+                raise TypeError(f"can't operate with object of type {type(other)}")
+            return other
+
+    def __add__(self, other):
```
bcebere, line 80: Do we need to keep the python arithmetics overloads in pybind in this case?

bcebere (approved):
This looks so much better, amazing work! LGTM

tenseal/tensors/abstract_tensor.py
youben11: Currently yes, because I'm usig the arithmetic operators (+,-,*) in the implementation, but yeah I can remove them and start using functions

Issue #192, opened 2020-12-15
Reshaping CKKSTensor


youben11:
## Feature Description
We want to reshape our tensors, without any memory copy, but using the strided view we already have internally.

## Affected Dependencies
This will have an impact on the currently implemented element-wise operations. They are only iterating over the internal storage, however, different shapes require a different iteration.

## Is your feature request related to a problem?
Future features might hardly depends on the shape of the tensor.

## Additional Context
Element-wise operations of different shape might need an iterator that traverses dimensions one by one in order to yield the same order of elements for element-wise operations.

bcebere:
I can work on this one.

PR #194, opened 2020-12-18
CKKSTensor reshaping

bcebere:
## Description
This PR adds support for CKKSTensor reshaping.

fixes https://github.com/OpenMined/TenSEAL/issues/192

## Checklist


tenseal/tensors/ckkstensor.py
```diff
@@ -87,3 +88,7 @@ def sum_batch(self) -> "CKKSTensor":
     def sum_batch_(self) -> "CKKSTensor":
         self.data.sum_batch_()
         return self
+
+    def reshape(self, shape: List[int]):
+        "Changes the internal representation to a new shape"
+        return self.data.reshape(shape)
```
youben11, line 94: I think reshape in most tensor library outputs a new tensor objects. Should we also have reshape (copy) and reshape_ (inplace) for this?

youben11: I'm not talking about the wrapper but the Cpp function

youben11 (changes requested):
Really great to have an underlying abstraction for the storage, well done Bogdan!
I've left few remarks regarding the API mainly.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -264,22 +273,22 @@ shared_ptr<CKKSTensor> CKKSTensor::op_plain_inplace(
         Plaintext plaintext;
         for (size_t i = start; i < end; i++) {
             this->tenseal_context()->encode<CKKSEncoder>(
-                operand_data[i], plaintext, this->_init_scale);
-            this->perform_plain_op(this->_data[i], plaintext, op);
+                operand_data.at(i), plaintext, this->_init_scale);
```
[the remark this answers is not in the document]
bcebere: operand_data is a vector here.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -273,7 +274,7 @@ shared_ptr<CKKSTensor> CKKSTensor::op_plain_inplace(
         Plaintext plaintext;
         for (size_t i = start; i < end; i++) {
             this->tenseal_context()->encode<CKKSEncoder>(
-                operand_data.at(i), plaintext, this->_init_scale);
+                operand_data[i], plaintext, this->_init_scale);
```
youben11, line 277: Can't we use the same API as for this->_data? namely `flat_at` on the operand instead of fetching all the data vector first? Just to make sure we are iterating over the same positions from both vectors, we are sure `flat_at` will return the same position for both of them, even if the implementation changes in the future.

youben11 (commented):
Thanks Bogdan! Just a last but not really important comment.

youben11 (approved):
Awesome!

PR #196, opened 2020-12-20
CKKSTensor Dot operation

### LINKED ISSUE #176: Dot product for CKKSTensor
## Feature Description

We should implement dot product depending on the tensor shape.
 - for 1D arrays.
 - for matrices.
 - for ND arrays(?)

youben11:
## Description
This PR adds the dot operation between a CKKSTensor and another, or a PlainTensor. The shapes supported now are 1D-1D (inner product) and 2D-2D (matrix multiplication).

#### TODO
- [x] Fix API
- [x] Parallel computation?

Fixes #176 

## How has this been tested?
- Added tests

## Checklist


tenseal/cpp/tensors/ckkstensor.h
```diff
@@ -63,6 +59,21 @@ class CKKSTensor : public EncryptedTensor<double, shared_ptr<CKKSTensor>>,
...
 
+    shared_ptr<CKKSTensor> dot_inplace(
+        const shared_ptr<CKKSTensor>& to_mul) override;
+    shared_ptr<CKKSTensor> dot_plain_inplace(
+        const PlainTensor<double>& to_mul) override;
+
+    shared_ptr<CKKSTensor> matmul(const shared_ptr<CKKSTensor> other) {
+        return this->copy()->matmul_inplace(other);
+    }
+    shared_ptr<CKKSTensor> matmul_inplace(const shared_ptr<CKKSTensor> other);
```
bcebere, line 70: There is something similar in the encrypted_vector interface. Maybe move them to the encrypted_tensor interface?

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -506,6 +494,91 @@ shared_ptr<CKKSTensor> CKKSTensor::polyval_inplace(
...
+    auto this_shape = this->shape();
+    auto other_shape = other->shape();
+
+    if (this_shape.size() == 1) {
+        if (other_shape.size() == 1) {  // 1D-1D
+            // inner product
+            this->mul_inplace(other);
+            this->sum_inplace();
+            return shared_from_this();
+        } else if (other_shape.size() == 2) {  // 1D-2D
```
bcebere, line 508: For broadcasting there are more checks needed: the dimensions must be equal or one of them = 1.

Maybe leave the condition 
```
if (this_shape.size() == 1 && other_shape.size() == 1)
```
and a TODO, if the broadcasting is not intended for this PR. Or throw an exception in this case too.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -506,6 +494,91 @@ shared_ptr<CKKSTensor> CKKSTensor::polyval_inplace(
...
+            // TODO: better implement broadcasting for mul first then would be
+            // implemented similar to 1D-1D
+        } else {
+            throw invalid_argument(
+                "don't support dot operations of more than 2 dimensions");
+        }
+    } else if (this_shape.size() == 2) {
+        if (other_shape.size() == 1) {  // 2D-1D
+            // TODO: better implement broadcasting for mul first then would be
```
bcebere, line 517: same as above

bcebere (changes requested):
Great work! 
just a few observations

tenseal/cpp/tensors/ckkstensor.cpp
youben11: it's actually just a placeholder for now and I would only need to send an error saying that this shapes aren't supported yet

tenseal/cpp/tensors/ckkstensor.h
youben11: True! This will require removing the n_jobs parameter from matmul, which I can't remember why I chose to put it there... I will open another PR for the API change after this one.

tenseal/cpp/tensors/ckkstensor.cpp
bcebere: still, it should throw an exception.

tenseal/cpp/tensors/ckkstensor.cpp
```diff
@@ -508,6 +496,223 @@ shared_ptr<CKKSTensor> CKKSTensor::polyval_inplace(
...
+            }
+            Ciphertext acc(*this->tenseal_context()->seal_context(),
+                           to_sum[0].parms_id());
+            evaluator->add_many(to_sum, acc);
+            // set element[row, col] to the computed inner product
+            new_data[i] = acc;
+        }
+        return true;
+    };
+
```
bcebere, line 685: we should somehow reuse the workers code, it is a lot of duplication here. 
but for another PR.

bcebere (approved):
Awesome work! LGTM

tenseal/cpp/tensors/ckkstensor.cpp
youben11: Yeah, I literally did a copy/paste here. I will open two refactor issues that I've found useful to work on

Issue #193, opened 2020-12-16
Lazy loading of TenSEALContext


youben11:
## Feature Description
When an Encrypted Tensor is deserialized, it need to be linked to a TenSEALContext to be operational, and that's why we require the context to be passed during deserialization. However, we can deserialize without the TenSEALContext, setting the tensor to be non functional, then requiring to link it to a specific context later to make it functional.

## Is your feature request related to a problem?
We often expect the serialized output to be deserialized as is, without requiring extra information.

bcebere:
I can work on this one.

PR #197, opened 2020-12-21
Lazy context

bcebere:
## Description
This PR introduces the capability of loading a protobuffer without having the context linked yet.
fixes https://github.com/OpenMined/TenSEAL/issues/193

## Checklist

### DIFF
--- a/tenseal/cpp/tensors/encrypted_tensor.h
+++ b/tenseal/cpp/tensors/encrypted_tensor.h
@@ -38,17 +38,22 @@ class EncryptedTensor {
     };
     virtual PlainTensor<plain_data_t> decrypt(
         const shared_ptr<SecretKey>& sk) const = 0;
-
     /**
      * Negates a EncryptedTensor<plain_data_t, encrypted_t>.
      **/
     encrypted_t negate() const { return this->copy()->negate_inplace(); };
-    virtual encrypted_t negate_inplace() = 0;
+    encrypted_t negate_inplace() {
+        this->_check_operation_sanity();
+        return negate_inplace_impl();
+    };
     /**
      * Compute the square of the EncryptedTensor<plain_data_t, encrypted_t>.
      **/
     encrypted_t square() const { return this->copy()->square_inplace(); };
-    virtual encrypted_t square_inplace() = 0;
+    encrypted_t square_inplace() {
+        this->_check_operation_sanity();
+        return square_inplace_impl();
+    };
     /**
      * Compute the power of the EncryptedTensor<plain_data_t, encrypted_t> with
      *minimal multiplication depth.
@@ -56,7 +61,10 @@ class EncryptedTensor {
     encrypted_t power(unsigned int power) const {
         return this->copy()->power_inplace(power);
     };
-    virtual encrypted_t power_inplace(unsigned int power) = 0;
+    encrypted_t power_inplace(unsigned int power) {
+        this->_check_operation_sanity();
+        return power_inplace_impl(power);
+    };
     /**
      * Encrypted evaluation function operates on two encrypted tensors and
      * returns a new EncryptedTensor<plain_data_t, encrypted_t>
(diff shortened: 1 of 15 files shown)


tenseal/cpp/tensors/encrypted_tensor.h
```diff
@@ -161,6 +161,10 @@ class EncryptedTensor {
      **/
     void link_tenseal_context(shared_ptr<TenSEALContext> ctx) {
         this->_context = ctx;
+        if (_lazy_buffer) {
+            this->load(*_lazy_buffer);
+            _lazy_buffer = {};
+        }
```
youben11, line 167: maybe have an else here which throw a weird error

youben11 (changes requested):
Nice and clean! I'm now wondering what happens when we run an operation which requires the tenseal_context? A way to deal with this is to always fetch the tenseal_context using the get method `tenseal_context()`, which will handle the case when the tensor is still lazy loaded, throwing an error message to the user. What do you think?

bcebere:
> Nice and clean! I'm now wondering what happens when we run an operation which requires the tenseal_context? A way to deal with this is to always fetch the tenseal_context using the get method `tenseal_context()`, which will handle the case when the tensor is still lazy loaded, throwing an error message to the user. What do you think?

There was another attempt in the PR, but I reverted, you can see here that I was going for a "call guardian" at first https://github.com/OpenMined/TenSEAL/pull/197/commits/b2533dc385471338a3eb4d4f808cfd08ee2cd333.
But the changes were too big for the gain.

Yeah, I will check to use tenseal_context everywhere, it is easier.

tenseal/cpp/tensors/encrypted_tensor.h
bcebere: The "else" case here is the standard call to "link_tenseal_context", used in constructors. we should not throw anything there. We just want to unpack the optional lazy_buffer, otherwise we will rely on the null context exceptions.

bcebere: For example, this line would go in the "else" branch https://github.com/OpenMined/TenSEAL/blob/master/tenseal/cpp/tensors/ckksvector.cpp#L11

youben11:
> > Nice and clean! I'm now wondering what happens when we run an operation which requires the tenseal_context? A way to deal with this is to always fetch the tenseal_context using the get method `tenseal_context()`, which will handle the case when the tensor is still lazy loaded, throwing an error message to the user. What do you think?
> 
> There was another attempt in the PR, but I reverted, you can see here that I was going for a "call guardian" at first [b2533dc](https://github.com/OpenMined/TenSEAL/commit/b2533dc385471338a3eb4d4f808cfd08ee2cd333).
> But the changes were too big for the gain.
> 
> Yeah, I will check to use tenseal_context everywhere, it is easier.

I think as far as it's clear that the tenseal context is a private attribute and shouldn't be used directly (we can mention that in the documentation => this makes me remember that I really need to start documenting my code :p ) we should be good with throwing an error on the getter.

bcebere:
> > > Nice and clean! I'm now wondering what happens when we run an operation which requires the tenseal_context? A way to deal with this is to always fetch the tenseal_context using the get method `tenseal_context()`, which will handle the case when the tensor is still lazy loaded, throwing an error message to the user. What do you think?
> > 
> > 
> > There was another attempt in the PR, but I reverted, you can see here that I was going for a "call guardian" at first [b2533dc](https://github.com/OpenMined/TenSEAL/commit/b2533dc385471338a3eb4d4f808cfd08ee2cd333).
> > But the changes were too big for the gain.
> > Yeah, I will check to use tenseal_context everywhere, it is easier.
> 
> I think as far as it's clear that the tenseal context is a private attribute and shouldn't be used directly (we can mention that in the documentation => this makes me remember that I really need to start documenting my code :p ) we should be good with throwing an error on the getter.

We can make the tenseal contexr `private` in the encrypted_tensor instead of `protected`, and this way we'd be sure that everyone is using it via tenseal_context(), even the derived classes.

tenseal/cpp/tensors/encrypted_tensor.h
youben11: Got it!

youben11:
Yeah, this would work perfectly I guess

bcebere:
> Yeah, this would work perfectly I guess

Actually, we already use tenseal_context in the derived classes, but we need to access _context for the lazy loading, instead of throwing an exception. So, there isn't much too do other than adding a guardian for all calls, to do sanity checks. But I think that is a bit overkill.

youben11:
Yeah, I also agree that we shouldn't go for the guardian thing. If there isn't any way to allow only the lazy loader to access that `private` attribute, then we should leave it protected but make it clear that getting that attribute is forbidden by the supreme court

bcebere:
> Yeah, I also agree that we shouldn't go for the guardian thing. If there isn't any way to allow only the lazy loader to access that `private` attribute, then we should leave it protected but make it clear that getting that attribute is forbidden by the supreme court

updated

youben11 (approved):
Even better this way! LGTM!

PR #210, opened 2021-01-13
Symmetric encryption support

bcebere:
## Description
We want to expose the symmetric encryption support in the TenSEAL context.
 - [x] extend the TenSEAL context for symmetric encryption.
 - [x] add gtests.
 - [x] add python tests.

## Checklist


tenseal/cpp/context/tensealcontext.h
```diff
@@ -12,6 +13,11 @@ namespace tenseal {
 using namespace seal;
 using namespace std;
 
+enum class encryption_type {
+    public_key,
+    symmetric,
+};
+
```
youben11, line 20: In terms of naming, I think we have to choose between private/public and asymmetric/symmetric. Personally, I like the second one, as it doesn't have overlapping terms with the different keys (e.g. public keys)

tenseal/cpp/context/tensealcontext.h
```diff
@@ -27,8 +33,6 @@ class TenSEALContext {
      * time for doing operations on encrypted tensors since we only need to
      * instantiate them once.
      **/
-    shared_ptr<Encryptor> encryptor;
-    shared_ptr<Decryptor> decryptor;
```
youben11, line 31: The reason that we kept these objects in the beginning was because we didn't want to recreate them everytime we needed them. I see that you dropped the decryptor, is there any reason for that?

tenseal/enc_context.py
```diff
@@ -86,6 +93,7 @@ def __init__(
 
         Args:
             scheme : define the scheme to be used, either SCHEME_TYPE.BFV or SCHEME_TYPE.CKKS.
+            encryption_type : define the encryption type to be used, either ENCRYPTION_TYPE.PUBLIC_KEY, or ENCRYPTION_TYPE.SYMMETRIC.
```
youben11, line 96: Better put the docs in the same order as the parameters (encryption_type after coeff)

youben11 (changes requested):
Great addition to TenSEAL! I've left few comments

tenseal/cpp/context/tensealcontext.h
bcebere: The decryptor wasn't used anywhere. We already recreate it from a secret key, local or provided.

youben11 (approved):
LGTM! I'm also having some timeout in the bazel tests on another PR, we may need to increase that later, but I think we can still merge it as is. Your call!

PR #213, opened 2021-01-16
Usability improvements: Arbitrary sized vectors support

bcebere:
## Description
This PR introduces support for vector of arbitrary sizes for encrypted vectors(BFVVector, CKKSVector).

Until now, the user has to split the input in <slot_count>-sized chunks, and serialize each encrypted vector.

The chunked-vector format will support only some basic operations.

## Checklist


youben11 (changes requested):
The feature itself is awesome! However, lots of methods, expect the object to contain a single ciphertext. For example, matmul relies heavily on ciphertext rotation, which isn't trivial to do in this case. So I expect a lot of those features to break when dealing with large vectors.

Can we instead create new tensor types, specifically for these use cases? Or somehow extend ckks and bfv vectors and drop all operations but only element wise operations, that's actually the ones not requiring rotations and which should work fine here.

tenseal/cpp/tensors/bfvvector.cpp
```diff
@@ -187,8 +227,17 @@ shared_ptr<BFVVector> BFVVector::dot_plain_inplace(
...
 
 shared_ptr<BFVVector> BFVVector::sum_inplace(size_t /*axis=0*/) {
-    sum_vector(this->tenseal_context(), this->_ciphertext, this->size());
-    this->_size = 1;
+    vector<Ciphertext> interm_sum;
+    for (size_t idx = 0; idx < this->_ciphertexts.size(); ++idx) {
+        Ciphertext out = this->_ciphertexts[idx];
+        sum_vector(this->tenseal_context(), out, this->_sizes[idx]);
+        interm_sum.push_back(out);
+    }
```
youben11, line 235: This can actually be computed in parallel. Please leave a TODO. Maybe after adding some helpers for parallel computations, then we can get back to this

tenseal/cpp/tensors/bfvvector.cpp
```diff
@@ -282,40 +345,51 @@ shared_ptr<BFVVector> BFVVector::enc_matmul_plain_inplace(
...
-        tmp = this->_ciphertext;
+        this->tenseal_context()->evaluator->add_inplace(this->_ciphertexts[0],
+                                                        tmp);
+        tmp = this->_ciphertexts[0];
     }
 
-    this->_size = n;
+    this->_sizes = {n};
     return shared_from_this();
 }
```
youben11, line 369: Can't we implement this on multi-ciphertext vectors? I was thinking of two possibilities:
1. n is bigger than available slots, then we can just fill the first ct with the first value, then replicate that ciphertext until having enough cts.
2. n is smaller than available slots, then we can manipulate the first ct as we are doing right now

tenseal/cpp/tensors/ckksvector.cpp
```diff
@@ -197,101 +242,135 @@ shared_ptr<CKKSVector> CKKSVector::dot_plain_inplace(const plain_t& to_mul) {
...
 
 shared_ptr<CKKSVector> CKKSVector::sum_inplace(size_t /*axis = 0*/) {
-    sum_vector(this->tenseal_context(), this->_ciphertext, this->size());
-    this->_size = 1;
+    vector<Ciphertext> interm_sum;
+    for (size_t idx = 0; idx < this->_ciphertexts.size(); ++idx) {
+        Ciphertext out = this->_ciphertexts[idx];
+        sum_vector(this->tenseal_context(), out, this->_sizes[idx]);
+        interm_sum.push_back(out);
+    }
```
youben11, line 250: Same here for parallel computation. Please just leave a TODO

tenseal/cpp/tensors/ckksvector.cpp
```diff
@@ -413,46 +494,60 @@ shared_ptr<CKKSVector> CKKSVector::enc_matmul_plain_inplace(
         this->add_inplace(tmp);
     }
 
-    this->_size = rows_nb;
```
youben11, line 416: I don't think this method can work for multi-ciphertext vectors. You can see the rotation on line 492 which shouldn't work correctly in that case.

tenseal/cpp/tensors/ckksvector.cpp
```diff
@@ -413,46 +494,60 @@ shared_ptr<CKKSVector> CKKSVector::enc_matmul_plain_inplace(
...
+    this->_sizes = {rows_nb};
 
     return shared_from_this();
 }
 
 shared_ptr<CKKSVector> CKKSVector::replicate_first_slot_inplace(size_t n) {
+    if (this->_ciphertexts.size() != 1)
+        throw invalid_argument("can't replicate first_slot on chunked vectors");
+
```
youben11, line 505: Same as BFVVector, I think this can be supported for multi-ciphertext vectors

tenseal/cpp/tensors/encrypted_vector.h
```diff
@@ -51,15 +51,21 @@ class EncryptedVector : public EncryptedTensor<plain_t, encrypted_t> {
...
      * Return information about the ciphertext.
      **/
-    size_t ciphertext_size() const { return this->_ciphertext.size(); }
-    const Ciphertext& ciphertext() const { return this->_ciphertext; }
-    void ciphertext(Ciphertext&& other) { this->_ciphertext = other; }
+    size_t ciphertext_size() const {
+        size_t res = 0;
+        for (auto& ct : this->_ciphertexts) res += ct.size();
+        return res;
+    }
```
youben11, line 66: Maybe return a vector of ct sizes here?

tenseal/cpp/tensors/plain_tensor.h
```diff
@@ -202,6 +202,17 @@ class PlainTensor {
      * Casts the tensor to an 1D vector.
      */
     operator vector<plain_t>() const { return _data.data(); }
+    /**
+     * Split the tensor in chunks
+     * */
+    vector<PlainTensor<plain_t>> chunks(size_t max_size) const {
+        auto storage_chunks = _data.chunks(max_size);
+
+        vector<PlainTensor<plain_t>> result;
```
youben11, line 211: maybe reserve here

tenseal/cpp/tensors/tensor_storage.h
```diff
@@ -330,6 +331,18 @@ class TensorStorage {
...
         _data = xt::adapt(flat_data, {flat_data.size()});
     }
+    /**
+     * Split the internal storage in 1D chunks of max_size maximum size.
+     * */
+    vector<TensorStorage<dtype_t>> chunks(size_t max_size) const {
+        auto flat_data = vector<dtype_t>(_data.begin(), _data.end());
+        auto storage_chunks = split_vector(flat_data, max_size);
+
+        vector<TensorStorage<dtype_t>> result;
```
youben11, line 341: maybe reserve here

tenseal/cpp/tensors/utils/utils.h
```diff
@@ -93,7 +93,7 @@ shared_ptr<T> pack_vectors(const vector<shared_ptr<T>>& vectors) {
     replicate_vector(replicated_mask, slot_count);
 
     auto packed_vec = vectors[0]->copy();
-    packed_vec->size(slot_count);
+    packed_vec->chunked_size({slot_count});
```
youben11, line 96: This function can actually be redesigned to support multi-ciphertext vectors. Not necessarily in this PR.

youben11 (changes requested):
Really cool feature to add! I've left few comments.

tenseal/binding.cpp
```diff
@@ -80,12 +81,16 @@ PYBIND11_MODULE(_tenseal_cpp, m) {
...
-            return BFVVector::Create(ctx, data);
-        }))
-        .def(py::init(
-            [](const shared_ptr<TenSEALContext> &ctx, const std::string &data) {
-                return BFVVector::Create(ctx, data);
-            }))
+                 return BFVVector::Create(ctx, data);
+             }),
+             py::call_guard<py::scoped_ostream_redirect,
+                            py::scoped_estream_redirect>())
```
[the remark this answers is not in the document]
bcebere: By default, pybind11 redirects all the logging done in c++ to /dev/null.
You must add this call_guard to redirect the logging to sys.stdout and sys.stderr.

Otherwise, it won't show the warnings for big vectors.

youben11 (approved):
LGTM! Except that `replicate_first_slot_inplace` need some updates to work with chunked vectors, I'm sure it won't pass tests if we add cases with large vectors. So we can either merge it by disabling it with chunked vectors temporarily, opening an issue to make the updates later. Or keep it open till we add those updates.

I would prefer the first. But feel free to do what you see fit.

PR #227, opened 2021-01-30
Tweaks&improvements

bcebere:
## Description
 - context - init keys to null.
 - use clang as the toolchain for Linux. Our benchmarks show that Clang is faster than GCC on Linux. On Windows and OSX the existing toolchains are faster.

## Checklist


youben11:
That's great! Can we also enable clang for the workflows? On my local machine, I'm just setting env variables CXX=/usr/bin/clang++ and CC=/usr/bin/clang (just in case) and it works with CMake.

bcebere:
> That's great! Can we also enable clang for the workflows? On my local machine, I'm just setting env variables CXX=/usr/bin/clang++ and CC=/usr/bin/clang (just in case) and it works with CMake.

Yes, it is in this commit https://github.com/OpenMined/TenSEAL/pull/227/commits/62b0be5a377c8f4efafd511e805c4d1f2e37b377

youben11 (approved):

PR #285, opened 2021-04-19
BFVTensor

comidan:
## Description
This PR tries to add support for BFVTensor.
Some of the tests are not passing yet, unfortunately.

## Details
I didn't created a more generic encrypted_tensor interface but kept the original one and made a new one, I didn't know what would you have preferred to do. (I would have written a more generic interface inherited by the two for CKKS and BFV to have less duplicated code).

## Checklist


cmake/tests.cmake
```diff
@@ -16,7 +16,8 @@ if(${BUILD_TEST})
       ${TENSEAL_TESTS_BASEDIR}/tensors/ckksvector_test.cpp
       ${TENSEAL_TESTS_BASEDIR}/tensors/ckkstensor_test.cpp
       ${TENSEAL_TESTS_BASEDIR}/tensors/plaintensor_test.cpp
-      ${TENSEAL_TESTS_BASEDIR}/tensors/bfvvector_test.cpp)
+      ${TENSEAL_TESTS_BASEDIR}/tensors/bfvvector_test.cpp
```
bcebere, line 19: Why the duplicate cpp file?

tenseal/cpp/tensors/BUILD
```diff
@@ -16,9 +17,11 @@ cc_library(
     hdrs = [
         "api.h",
         "bfvvector.h",
+        "bfvtensor.h",
         "ckkstensor.h",
         "ckksvector.h",
         "encrypted_tensor.h",
+        "encrypted_tensor_bfv.h",
```
bcebere, line 24: encrypted_tensor should be a generic interface for all tensors in TenSEAL? Why do we need encrypted_tensor_bfv too?

tenseal/cpp/tensors/encrypted_tensor_bfv.h
```diff
@@ -0,0 +1,263 @@
+#ifndef TENSEAL_TENSOR_ENCRYPTED_TENSOR_BFV_H
```
bcebere, line 1: Why do we need a separate interface from encrypted_tensor?

tenseal/proto/tensors.proto
```diff
@@ -9,6 +9,16 @@ message BFVVectorProto {
...
 };
 
+//BFVTensor parameters
+message BFVTensorProto {
+    // The shape of the encrypted tensor
+    repeated uint32 shape = 1;
+    // The serialized ciphertexts
+    repeated bytes ciphertexts = 2;
+    // Optional batch size. Exists only if batching is enabled
+    uint32 batch_size = 4;
```
bcebere, line 19: Why 4? this can be index 3.

tests/cpp/tensors/bfvtensor_test.cpp
```diff
@@ -0,0 +1,532 @@
...
+#include "tenseal/cpp/tenseal.h"
+
+namespace tenseal {
+namespace {
+
+using namespace ::testing;
+using namespace std;
+
+template <class Iterable>
+bool are_close(const Iterable& l, const std::vector<int64_t>& r) {
```
bcebere, line 12: `are_close` is relevant only for CKKS. For BFV, you should get the exact result. See the BFVVector C++ tests for reference

tests/python/tenseal/tensors/test_bfv_tensor.py
```diff
@@ -0,0 +1,771 @@
...
+
+BROADCAST_SHAPES = [
+    ([2, 3], [2, 1]),
+    ([2, 1], [2, 4]),
+    ([2, 2], [4, 2, 2]),
+    ([2, 2, 3, 4, 5], [2, 3, 4, 5]),
+]
+
+
+def _almost_equal_number(v1, v2, m_pow_ten):
```
bcebere, line 31: Similarly, `_almost_equal_number` is only for CKKS. Here you should test for the exact result. See the test_bfv_vector tests for reference.

bcebere (changes requested):
Thank you for the contribution. It is a major step in the right direction!

One significant problem is that we should reuse the encrypted_tensor interface, which should be the root class for BFVTensor.
There is also a bit of code duplication between CKKSTensor and BFVTensor, but we can think in another PR about templating the classes.


Regarding the failing tests, it looks like a linting problem. Make sure that you run 
```.github/workflows/scripts/lint_cpp.sh```
and
```.github/workflows/scripts/lint_python.sh```

and they are successful before pushing. You might need to install `clang-format,` `flake8` and `black` locally for that.

Other than that, awesome work so far!

File: cmake/tests.cmake (hunk 1 above)
Comment (comidan) [reply cmake/tests.cmake:19]: No sorry probably I just copy pasted and changed again after having saved, those lines should be appearing:
```cmake
${TENSEAL_TESTS_BASEDIR}/tensors/bfvvector_test.cpp
${TENSEAL_TESTS_BASEDIR}/tensors/bfvtensor_test.cpp)
```

tenseal/cpp/tensors/BUILD
comidan: Yes you are right of course, sorry.
I just got a little confused with the _scale_ method in _encrypted_tensor.h_ given that it is not supposed to be used in BFV and indeed only now I've discovered that it's declared as not implemented for BFV (for the BFVVector counterpart I mean).
Also the same for _polyval_inplace_ method which was using double so I initially changed the data type to the template _plain_data_t_ but it was messing with BFVVector which was using double as template but the method itself was again declared as not implemented, so I did the same now for BFVTensor.

Sorry for the confusion

tenseal/cpp/tensors/encrypted_tensor_bfv.h
comidan: No we don't sorry again, the wrong reason was the one mentioned above

tenseal/proto/tensors.proto
comidan: I should stop copy-pasting to be faster and forgetting about it, sorry.
I deleted the scale which was not needed but forgot indeed to change the index to 3.

tests/cpp/tensors/bfvtensor_test.cpp
comidan: Right thank you, I changed it now to assert/expect the exact result indeed

tests/python/tenseal/tensors/test_bfv_tensor.py
comidan: Yes right, thank you

comidan:
> Thank you for the contribution. It is a major step in the right direction!
> 
> One significant problem is that we should reuse the encrypted_tensor interface, which should be the root class for BFVTensor.
> There is also a bit of code duplication between CKKSTensor and BFVTensor, but we can think in another PR about templating the classes.
> 
> Regarding the failing tests, it looks like a linting problem. Make sure that you run
> `.github/workflows/scripts/lint_cpp.sh`
> and
> `.github/workflows/scripts/lint_python.sh`
> 
> and they are successful before pushing. You might need to install `clang-format,` `flake8` and `black` locally for that.
> 
> Other than that, awesome work so far!

Ok thank you for the detailed explanation.
I run both of the scripts and it does not seem to give me any kind of errors (I run them after compiling and building everything).
The only output of the first script is the diff output from git, as the script indeed does after calling clang-format.
While the second one it tells me "All done!".

So I think it's ok and I'll proceed to push the changes.

bcebere:
> > Thank you for the contribution. It is a major step in the right direction!
> > One significant problem is that we should reuse the encrypted_tensor interface, which should be the root class for BFVTensor.
> > There is also a bit of code duplication between CKKSTensor and BFVTensor, but we can think in another PR about templating the classes.
> > Regarding the failing tests, it looks like a linting problem. Make sure that you run
> > `.github/workflows/scripts/lint_cpp.sh`
> > and
> > `.github/workflows/scripts/lint_python.sh`
> > and they are successful before pushing. You might need to install `clang-format,` `flake8` and `black` locally for that.
> > Other than that, awesome work so far!
> 
> Ok thank you for the detailed explanation.
> I run both of the scripts and it does not seem to give me any kind of errors (I run them after compiling and building everything).
> The only output of the first script is the diff output from git, as the script indeed does after calling clang-format.
> While the second one it tells me "All done!".
> 
> So I think it's ok and I'll proceed to push the changes.

The diff you are seeing contains the actual linting errors that break the build.

But my bad, it can be a bit confusing:
 - lint_cpp.sh runs clang-format, you should commit the diff, if any.
 - lint_python.sh does a dry-run, but it doesn't change anything. If lint_python.sh outputs anything at all, just run in the root of the repository `black .` to lint the py scripts.

comidan:
> > > Thank you for the contribution. It is a major step in the right direction!
> > > One significant problem is that we should reuse the encrypted_tensor interface, which should be the root class for BFVTensor.
> > > There is also a bit of code duplication between CKKSTensor and BFVTensor, but we can think in another PR about templating the classes.
> > > Regarding the failing tests, it looks like a linting problem. Make sure that you run
> > > `.github/workflows/scripts/lint_cpp.sh`
> > > and
> > > `.github/workflows/scripts/lint_python.sh`
> > > and they are successful before pushing. You might need to install `clang-format,` `flake8` and `black` locally for that.
> > > Other than that, awesome work so far!
> > 
> > 
> > Ok thank you for the detailed explanation.
> > I run both of the scripts and it does not seem to give me any kind of errors (I run them after compiling and building everything).
> > The only output of the first script is the diff output from git, as the script indeed does after calling clang-format.
> > While the second one it tells me "All done!".
> > So I think it's ok and I'll proceed to push the changes.
> 
> The diff you are seeing contains the actual linting errors that break the build.
> 
> But my bad, it can be a bit confusing:
> 
> * lint_cpp.sh runs clang-format inplace, you should commit the diff, if any.
> * lint_python.sh does a dry-run, but it doesn't change anything. If lint_python.sh outputs anything at all, just run in the root of the repository `black .` to lint the py scripts.

Ok thank you, I run `black .` at the root of the repository and it considered two more files than the ones considered by the script to reformat. It also prints at then end of the log "All done!".
Regarding the diff, ok I'll then commit them!

comidan:
Ok now with the new commit it shows that by running the python tests some of them pass but then it stops at one giving indeed the same error I get locally: `ValueError: SEAL: couldn't find context_data from params_id`,

bcebere:
> Ok now with the new commit it shows that by running the python tests some of them pass but then it stops at one giving indeed the same error I get locally: `ValueError: SEAL: couldn't find context_data from params_id`,

I pushed some fixes, you can see the summary from the commit.
 - you don't need rescaling or coeff modulus for BFV
 - BFV doesn't need modulus switch, that's why you saw that error. 
 - the C++ tests weren''t compiling because you were using vector<int> instead of vector<int64_t>
 - 
I haven't checked the python tests, though/

comidan:
> > Ok now with the new commit it shows that by running the python tests some of them pass but then it stops at one giving indeed the same error I get locally: `ValueError: SEAL: couldn't find context_data from params_id`,
> 
> I pushed some fixes, you can see the summary from the commit.
>  - you don't need rescaling or coeff modulus for BFV
>  - BFV doesn't need modulus switch, that's why you saw that error. 
>  - the C++ tests weren't compiling because you were using vector<int> instead of vector<int64_t>
>  - 
> I haven't checked the python tests, though.

Thank you!

Yes my bad, I indeed removed rescaling and coefficient modulus from various methods but not everywhere, sorry for the confusion.

Ok I'll wait for the tests to run here on GitHub and then I'll - further check the python for problems, if any.

comidan:
I pushed fixes to the python tests of BFVTensor, now locally they all work! Should also then work everything here in theory.
There were few issues with the shapes passed to numpy _randint_ and some tests thought to be for CKKS polyval which did not make sense here.

tenseal/cpp/tensors/bfvtensor.h
```diff
@@ -0,0 +1,193 @@
...
+
+using namespace seal;
+using namespace std;
+
+class BFVTensor : public EncryptedTensor<int64_t, shared_ptr<BFVTensor>>,
+                  public enable_shared_from_this<BFVTensor> {
+   public:
+    using EncryptedTensor<int64_t, shared_ptr<BFVTensor>>::decrypt;
+    /**
+     * Create a new BFVTensor from an 1D vector.
```
bcebere, line 18: The comment here is not correct, the BFVTensor can be created from tensors of any shape

bcebere (approved):
LGTM! Great work! Thank you for your contribution, @comidan!

I've left just one minor observation, after that we can merge the PR.

tenseal/cpp/tensors/bfvtensor.h
comidan: > LGTM! Great work! Thank you for your contribution, @comidan!
> 
> I've left just one minor observation, after that we can merge the PR.

Thank you for the help!

Sure ok, I'll just pushed the new header file with the fixed comment.

tenseal/cpp/tensors/bfvtensor.h
```diff
@@ -0,0 +1,193 @@
...
+class BFVTensor : public EncryptedTensor<int64_t, shared_ptr<BFVTensor>>,
+                  public enable_shared_from_this<BFVTensor> {
+   public:
+    using EncryptedTensor<int64_t, shared_ptr<BFVTensor>>::decrypt;
+    /**
+     * Create a new BFVTensor from a tensor of any shape.
+     * @param[in] input vector.
+     * @param[in] input vector.
+     * @param[in] input vector.
+     */
```
bcebere, line 22: Sorry, but the comments still don't make much sense.

Why the
```  * @param[in] input vector.
     * @param[in] input vector.
     * @param[in] input vector.```
     ?

comidan: Ok @bcebere I'll change it accordingly, I just left the same comments which were already present in [CKKSTensor.h](https://github.com/OpenMined/TenSEAL/blob/1b371b678c804c47a993caee9088508b6a2bd1f5/tenseal/cpp/tensors/ckkstensor.h#L18) assuming they were correct.

Should I then change also the ones for CKKSTensor?

bcebere: Yes, please. Thank you so much for improving this!

bcebere:
Thank you for your contribution, @comidan!

comidan:
> Thank you for your contribution, @comidan!

Thank you for the help in making it!

Issue #398, opened 2022-05-23
ckksvector multiplied with identity matrix results in transparent ciphertext


MartinNoc:
## Question
Is it a bug or intended behavior when I get a "result ciphertext is transparent" error when multiplying a ckksvector with a numpy identity matrix?

## Further Information
When multiplying a ckksvector with a numpy identity matrix, I get the error "result ciphertext is transparent" which seems very strange to me, as the result should be the same ciphertext as before, right?
Moreover, when I compute the dot product of the ciphertext and any identity matrix column, it works.

## System Information
 - OS: Windows
 - Python 3.9.7, tenseal 0.3.8
 - Conda 4.11.0

## Code
```python
import tenseal as ts
import numpy as np
  
# Create TenSEAL context
bits_scale = 32
context = ts.context(
    ts.SCHEME_TYPE.CKKS,
    poly_modulus_degree=16384,
    coeff_mod_bit_sizes=[43] + [bits_scale]*11 + [43]
)
context.global_scale = 2**bits_scale
context.generate_galois_keys()

N = 100
x = np.random.rand(N)
eye = np.eye(N)
x_enc = ts.ckks_vector(context, x)

# this works
x_enc.dot(eye[:,0])

# this results in an error
x_enc @ eye
```

MartinNoc:
Ok, I figured out, TenSEAL uses the diagonal method to compute vector-matrix products. This means for the identity matrix that all vectors except the first one are zero, resulting in the transparent ciphertext error.

[See here](https://github.com/OpenMined/TenSEAL/blob/ec21905822854291e6b93b5641c59bebd14266a7/tenseal/cpp/tensors/encrypted_vector.h#L151) the corresponding code piece.

How can we solve this?
In my opinion, the transparent ciphertext should be thrown for the resulting product, not for intermediate results.

PR #405, opened 2022-06-20
ignoring zero diagonals in diagonal_ct_vector_matmul

mrader1248:
## Description
As explained in #398, matrices with diagonals that contain nothing but zeros cannot be multiplied with a vector, as this results in `ValueError: result ciphertext is transparent`. This can be fixed by simply ignoring zero diagonals, which in addition potentially saves some unnecessary computations.

## Affected Dependencies
None

## How has this been tested?
I used the following code to test, which previously resulted in the error explained above.
```python
import numpy as np
import tenseal as ts

n = 1000
x = np.random.rand(n)
i = np.eye(n)
poly_mod_degree = 4096
coeff_mod_bit_sizes = [40, 20, 40]
ctx = ts.context(ts.SCHEME_TYPE.CKKS, poly_mod_degree, -1, coeff_mod_bit_sizes)
ctx.generate_galois_keys()
ctx.global_scale = 2 ** 20
x_enc = ts.ckks_vector(ctx, x)
x_enc @ i
```

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,16 +57,18 @@ class CMakeBuild(build_ext):
         cfg = "Debug" if self.debug else "Release"
         build_args = ["--config", cfg]
 
+        env = os.environ.copy()
+
         if platform.system() == "Windows":
             cmake_args += [f"-DCMAKE_LIBRARY_OUTPUT_DIRECTORY_{cfg.upper()}={extdir}"]
-            if sys.maxsize > 2 ** 32:
+            if sys.maxsize > 2**32:
                 cmake_args += ["-A", "x64"]
             build_args += ["--", "/m", "/p:TrackFileAccess=false"]
         else:
             cmake_args += ["-DCMAKE_BUILD_TYPE=" + cfg]
-            build_args += ["--", "-j"]
+            if "MAKEFLAGS" not in env:
+                build_args += ["--", "-j"]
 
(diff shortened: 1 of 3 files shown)


mrader1248:
Is it possible that the test runner had a problem or `_almost_equal` had a bad day? I cannot reproduce the reported failure of
```
tests/python/tenseal/tensors/test_ckks_tensor.py::test_subscript[data0-slices0-new_shape0]
```
locally.

mrader1248:
@bcebere, could you help me here?

mrader1248:
If `make` is called with `-j` the number of processes for the the build process is not limited and it crashes unless there is enough RAM. For this reason, I modified `setup.py` such that `-j` is only added if the environment variable `MAKEFLAGS` is not set. With this, `setup.py` respects a user's default make flags, e.g. defined in `.zshrc` or `.bashrc`, or `MAKEFLAGS="-j4" pip install ...` can be called to limit the number of processes directly. The default behaviour is unchanged (unlimited number of processes if `MAKEFLAGS` is not set).

MartinNoc:
@youben11 
Does anybody have time to review this PR? Thanks 🙏

bcebere:
LGTM!
Thank you for your contribution, @mrader1248 !

PR #521, opened 2026-07-29
build: modernize Python packaging and linting (PEP 621, ruff, uv)

philomath213:
Replaces the flake8/black lint stack with ruff + pre-commit, and moves packaging onto current standards.

- **ruff replaces flake8 and black**, run through pre-commit (which also owns clang-format); the Linter job moves to Python 3.14
- **PEP 517/621**: adds the missing `[build-system]` table — builds currently rely on pip's legacy fallback — moves metadata to `[project]`, drops the `distutils.version.LooseVersion` import, adds `MANIFEST.in`
- **Dev dependencies via uv**: PEP 735 `[dependency-groups]` + `uv.lock`; removes `requirements_dev.txt`, `requirements_dev.lock` (it pinned a local `--find-links` path and was unusable) and `setup.cfg`
- **Declares `numpy` as a runtime dependency** — it was never declared, so a bare `pip install tenseal` produced an unimportable package
- Removes two dead CI scripts

Dropping black also closes both open Dependabot alerts, and is what unblocks linting on Python 3.14.

Verified: wheel builds end-to-end and imports, sdist carries `CMakeLists.txt`/`cmake/`/`.proto`, `twine check` and `pre-commit run --all-files` clean.

The `style:` commit is a mechanical formatting pass, split out to keep the tooling diff reviewable.


.github/workflows/scripts/install_req_docker.sh
```diff
@@ -13,7 +13,7 @@ wget https://github.com/Kitware/CMake/releases/download/v3.17.0/${cmake_install}
 sh ${cmake_install} --skip-license --prefix=/usr/local/ --exclude-subdir
 
 python -m pip install --upgrade pip
-pip install -r requirements_dev.txt
+pip install --group dev
```
youben11, line 16: Shouldn't we use uv? pip won't use the lock file AFAIK. If the setup is based on uv, then we probably should use it instead of pip everywhere. Note: uv have a pip subcommand

youben11 (approved):
I'm not sure about the use of uv everywhere, but that's what I would expect it to be if there is a lock file. Not really important for now, so I approve and let you decide.

philomath213:
> I'm not sure about the use of uv everywhere, but that's what I would expect it to be if there is a lock file. Not really important for now, so I approve and let you decide.

Good catch. I have introduced UV to make it easy for the end users to stay on an older Python without fighting system interpreters.

The lock file is committed, but nothing actually reads it yet.
I’ll do this in a follow-up PR, since it changes benchmarks.yml and #520 already touches that file. Benchmarks is also where the lock helps most: if torch or numpy quietly changes version, the timings move and we can no longer compare runs.

